PR #1, opened 2021-02-24
storage node proposal

kelvich:
[Rendered](https://github.com/libzenith/rfcs/blob/storage_design/002-storage.md)

### DIFF
--- a/002-storage.md
+++ b/002-storage.md
@@ -1,0 +1,186 @@
+# Storage fleet
+
+# **Storage fleet**
+
+- When database is smaller then a chunk size we naturally can store them in one chunk (since their page_key would fit in some chunk's [hi, lo) range).
+
+<img width="937" alt="Screenshot_2021-02-22_at_16 49 17" src="https://user-images.githubusercontent.com/284219/108729836-ffcbd200-753b-11eb-9412-db802ec30021.png">
+
+Few databases are stored in one chunk, replicated three times
+
+- When database can't fit into one storage node it can occupy lots of chunks that were split while database was growing. Chunk placement on nodes is controlled by us with some automatization, but we alway may manually move chunks around the cluster.
+
+<img width="940" alt="Screenshot_2021-02-22_at_16 49 10" src="https://user-images.githubusercontent.com/284219/108729815-fb071e00-753b-11eb-86e0-be6703e47d82.png">
+
+Here one big database occupies two set of nodes. Also some chunks were moved around to restore replication factor after disk failure. In this case we also have "sharded" storage for a big database and issue wal writes to different chunks in parallel.
+
+## **Chunk placement strategies**
+
+There are few scenarios where we may want to move chunks around the cluster:
+
+- disk usage on some node is big
+- some disk experienced a failure
+- some node experienced a failure or need maintenance
+
+## **Chunk replication**
+
+Chunk replication may be done by cloning page ranges with respect to some lsn from peer nodes, updating global metadata, waiting for WAL to come, replaying previous WAL and becoming online -- more or less like during chunk split.
+
+
+# Zenith storage node — alternative
+
+## **Design considerations**
+
+Simplify storage operations for people => Gain adoption/installs on laptops and small private installation => Attract customers to DBaaS by seamless integration between our tooling and cloud.
+
+Proposed architecture addresses:
+
+- High availability -- tolerates n/2 - 1 failures
+- Multi-tenancy -- one storage for all databases
+- Elasticity -- increase storage size on the go by adding nodes
+- Snapshots / backups / PITR with S3 offload
+- Compression
+
+Minuses are:
+
+- Quite a lot of work
+- Single page access may touch few disk pages
+- Some bloat in data — may slowdown sequential scans
+
+## **Summary**
+
+Storage cluster is sharded key-value store with ordered keys. Key (****page_key****) is a tuple of `(pg_id, db_id, timeline_id, rel_id, forkno, segno, pageno, lsn)`. Value is either page or page diff/wal. Each chunk (chunk == shard) stores approx 50-100GB ~~and automatically splits in half when grows bigger then soft 100GB limit~~. by having a fixed range of pageno's it is responsible for. Chunks placement on storage nodes is stored in a separate metadata service, so chunk can be freely moved around the cluster if it is need. Chunk itself is a filesystem directory with following sub directories:
+
+```
+
+|-chunk_42/
+  |-store/ -- contains lsm with pages/pagediffs ranging from
+  |	      page_key_lo to page_key_hi
+  |-wal/
+  |  |- db_1234/ db-specific wal files with pages from page_key_lo
+  |		 to page_key_hi
+  |
+  |-chunk.meta -- small file with snapshot references
+		  (page_key_prefix+lsn+name)
+		  and PITR regions (page_key_start, page_key_end)
+```
+
+## **Chunk**
+
+Chunk is responsible for storing pages potentially from different databases and relations. Each page is addressed by a lexicographically ordered tuple (****page_key****) with following fields:
+
+- `pg_id` -- unique id of given postgres instance (or postgres cluster as it is called in postgres docs)
+- `db_id` -- database that was created by 'CREATE DATABASE' in a given postgres instance
+- `db_timeline` -- used to create Copy-on-Write instances from snapshots, described later
+- `rel_id` -- tuple of (relation_id, 0) for tables and (indexed_relation_id, rel_id) for indices. Done this way so table indices were closer to table itself on our global key space.
+- `(forkno, segno, pageno)` -- page coordinates in postgres data files
+- `lsn_timeline` -- postgres feature, increments when PITR was done.
+- `lsn` -- lsn of current page version.
+
+Chunk stores pages and page diffs ranging from page_key_lo to page_key_hi. Processing node looks at page in wal record and sends record to a chunk responsible for this page range. When wal record arrives to a chunk it is initially stored in `chunk_id/wal/db_id/wal_segno.wal`. Then background process moves records from that wal files to the lsm tree in `chunk_id/store`. Or, more precisely, wal records would be materialized into lsm memtable and when that memtable is flushed to SSTable on disk we may trim the wal. That way some not durably (in the distributed sense) committed pages may enter the tree -- here we rely on processing node behavior: page request from processing node should contain proper lsm horizons so that storage node may respond with proper page version.
+
+LSM here is a usual LSM for variable-length values: at first data is stored in memory (we hold incoming wal records to be able to regenerate it after restart) at some balanced tree. When this tree grows big enough we dump it into disk file (SSTable) sorting records by key. Then SStables are mergesorted in the background to a different files. All file operation are sequential and do not require WAL for durability.
+
+Content of SSTable can be following:
+
+```jsx
+(pg_id, db_id, ... , pageno=42, lsn=100) (full 8k page data)
+(pg_id, db_id, ... , pageno=42, lsn=150) (per-page diff)
+(pg_id, db_id, ... , pageno=42, lsn=180) (per-page diff)
+(pg_id, db_id, ... , pageno=42, lsn=200) (per-page diff)
+(pg_id, db_id, ... , pageno=42, lsn=220) (full 8k page data)
+(pg_id, db_id, ... , pageno=42, lsn=250) (per-page diff)
+(pg_id, db_id, ... , pageno=42, lsn=270) (per-page diff)
+(pg_id, db_id, ... , pageno=5000, lsn=100) (full 8k page data)
+```
+
+So query for `pageno=42 up to lsn=260` would need to find closest entry less then this key, iterate back to the latest full page and iterate forward to apply diffs. How often page is materialized in lsn-version sequence is up to us -- let's say each 5th version should be a full page.
+
+### **Page deletion**
+
+To delete old pages we insert blind deletion marker `(pg_id, db_id, #trim_lsn < 150)` into a lsm tree. During merges such marker would indicate that all pages with smaller lsn should be discarded. Delete marker will travel down the tree levels hierarchy until it reaches last level. In non-PITR scenario where old page version are not needed at all such deletion marker would (in average) prevent old page versions propagation down the tree -- so all bloat would concentrate at higher tree layers without affecting bigger bottom layers.
+
+### **Recovery**
+
+Upon storage node restart recent WAL files are applied to appropriate pages and resulting pages stored in lsm memtable. So this should be fast since we are not writing anything to disk.
+
+### **Checkpointing**
+
+No such mechanism is needed. Or we may look at the storage node as at kind of continuous chekpointer.
+
+### **Full page writes (torn page protection)**
+
+Storage node never updates individual pages, only merges SSTable, so torn pages is not an issue.
+
+### **Snapshot**
+
+That is the part that I like about this design -- snapshot creation is instant and cheap operation that can have flexible granularity level: whole instance, database, table. Snapshot creation inserts a record in `chunk.meta` file with lsn of this snapshot and key prefix `(pg_id, db_id, db_timeline, rel_id, *)` that prohibits pages deletion within this range. Storage node may not know anything about page internals, but by changing number of fields in our prefix we may change snapshot granularity.
+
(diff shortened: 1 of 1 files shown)


knizhnik:
First of all we need to determine the meaning of chunk. It is exchange unit between computation and storage nodes? Or it is unit if data partitioning between storage nodes? Or it is mostly physical area used to allocate space with minimal fragmentation at storage nodes?

Using Postgres 1G relation segment as chunk  seems to be the easiest solution compatible with vanilla Postgres. If we are replacing smgr layer, then  preserving 1Gb segment size (and even using fixed size segments is not so  critical for us.

What are the advantages of 1Gb segments? 
- It will be very inefficient if computation node will request individual pages.
- We do not need to implement some kind of prefetching, at least for sequential search

If chunk is mostly exchange unit, then having large enough chunks leads to
faster recovery and launching of new nodes.

When do we need to send chunks from storage nodes to computation nodes?
1. To launch new node 
2. To request missing chunks at master (when database doesn't fir in node's cache)
3. To update replica

In case 1 the large chunk size is the faster node will be started.
In case 2 large  chunks are also useful for  prefetch. 
In 3 case large chunks can cause sending a lot of redundant data.

I assume that for normal work processing of OLTP queries computation node need to have working set which in most cases will be equal to the whole database instance.
So we should provide fastest transfer of the whole database from storage to computation nodes (unlike Snowflake which we need to transfer data only needed for particular query). So it is "pros" for larger chunks.

Concerning updates of replicas snapshots,  large  chunks may be not so efficient: to update one page we need to transfer 1Gb of data. But what are the alternatives? Stats told, that YDB knows which data will be needed for transactions and request all this data from storage nodes. But I do not understand how we can get this knowledge from standard Postgres backend. So there are still three choices:
1. Request individual pages
2. Request chunks (size has to be discussed, but significantly larger than 8Kb)
3. Push updates to the replica, so it is not replica requesting invalidated pages, but storage nodes, which knows for sure what was changed, pushing this data to replica.

Thinking more about 3), leads to  another question: we are sending WAL rather then data pages from master to storage nodes because we assume it is much more efficient: allows to reduce network traffic. 
But why when we are sending data in opposite direction: from storage nodes to replicas we are not following the same approach? Yes, we we launch new replica, then definitely it is not efficient to stream all WAL to replica and make replica replay it. We should send some snapshot.  But if we are talking about updating replica's snapshot, will it be more efficient to send WAL pages and let replica replay them rather than send data pages constructed by storage nodes?

There seems to be some assymetry in Aurora model, doesn't it?

kelvich:
> Some my thought concerning chunks and multi-tenancy: (sorry for log of word, may be it is better to discuss it orally...
> First of all we need to determine the meaning of chunk. It is exchange unit between computation and storage nodes? Or it is unit if data partitioning between storage nodes? Or it is mostly physical area used allocate space with minimal fragmentation of storage nodes?

The way I used term chunk it is "unit if data partitioning between storage nodes" which is phycally lsm + some WAL files. Exchange unit is a page or stream of pages.

> Using Postgres 1G relation segment as chunk  seems to be the easiest solution compatible with vanilla Postgres. If we are replacing smgr layer, then  preserving 1Gb segment size (and even using fixed size segments is not so  critical for us.
> What are the advantages of 1Gb segments? 
> - It will be very inefficient if computation node will request individual pages.

For full scans we have strategy info and can request stream of pages. For index scan if we are hitting random pages downloading whole 1gb would be quite bad.

> - We do not need to implement some kind of prefetching, at least for sequential search

Making prefetch work for sequential scans seems easy for me -- we know in advance that it is sequential scan and we may request all pages.

> I assume that for normal work processing of OLTP queries computation node need to have working set which in most cases will be equal to the whole database instance. So we should provide fastest transfer of the whole database from storage to computation nodes (unlike Snowflake which we need to transfer data only needed for particular query).

And in some cases it wouldn't. Look at all this companies having 20PB postgres clusters for oltp. And if in that cases we provide orders of magnitude worse perfomance it will be bad.

> Finally few words about multi-tenancy

I meant that processing node is standalone postgres, but storage node software may serve lots of different processing nodes. Having separate storage node process for each processing node seems expensive and we won't have a global page cache in this situation.

knizhnik:
If I correctly understand your proposal, chunks will contain pages updated by most recent transactions.
Assume that chunk size is just two pages.
If transaction T1 with LSN1 touches pages P1 and P10, then we form chunk {P1(LSN1},P10(LSN1)}.
If transaction T@ with LSN2 touches pages P2 and P10, then we form chunk {P2(LSN2},P10(LSN2)}.
If we now want to launch new nodes, then we need to read both this chunks and form parcel {P1(LSN1},P10(LSN2)}.

It seems to be complex... and not so efficient.

From my point of view splitting relation in fixed size chunks (1Gb or whatever else) works nice for scattering data between storage nodes. So if we are treating chunk as "data partitionig unit". The same chunks can be used for snapshots and as unit of sending snapshots to computation nodes.

It is not efficient for refreshing replica. But in this case the most efficient format is ... WAL, isn't it?
If we are sending WAL from master to storage nodes, why should not we do the same when sending data from storage nodes to replicas? The original schema: send invalidations and let replica request pages itself seems to just adds unmotivated complexity...

knizhnik:
> And in some cases it wouldn't. Look at all this companies having 20PB postgres clusters for oltp. And if in that cases we provide orders of magnitude worse performance it will be bad.

Well, I based on the assumption that OLTP workload means permanent and more or less random access to all database data.
Unlike OLAP, when data analytic may once per day execute some complex query which traverse substantial part of data (but not all data),  in case of OLTP we expect to have constant stream of small queries which access few random records.

So it makes me think that we are talking about OLTP workload, then it is not possible to "swap out" some rarely used databases or just parts of this databases. From my point of view OLTP can normally work only if master's local cache contains all working set. For most databases this working set will be equal to the whole database. 

There is some contradiction with our goal to be able to support work with arbitrary large databases which doesn't fit in one node. But...
- As far as I understand our business model will be first of all oriented on small customers. We are not expecting that BofA will move it's database to the cloud.
- I do not believe that we can provide reasonable OLTP performance (and especially latencies) if all affected data will not be available locally. With only one major exception: databases containing historical data. In most cases applications are working with most recent data, but sometime (rarely) them need to access old data, but speed of such queries is not so critical. 
This case can be quite efficiently handled at application level using Postgres partitioning mechanism and using relation-segment distribution at storage level.

hlinnaka:
> It is not efficient for refreshing replica. But in this case the most efficient format is ... WAL, isn't it?
If we are sending WAL from master to storage nodes, why should not we do the same when sending data from storage nodes to replicas? The original schema: send invalidations and let replica request pages itself seems to just adds unmotivated complexity...

Yeah, the WAL is certainly the most straightforward format. Let's start with that. 

In the future, we can optimize: instead of sending the whole WAL, the read-only node can "subscribe" to receive only WAL that's applicable to the pages that the node has in cache. Or, only send an invalidation stream, not containing the actual changes, but just the page # of modified pages. Or, if a replica is jumping from LSN X to (LSN X+ 1000), don't stream all the individual WAL records in that range, but a combined "diff" between X and X+1000.

But let's start with just the raw WAL stream.

002-storage.md
```diff
@@ -0,0 +1,186 @@
+# Storage fleet
+
+# **Storage fleet**
```
adsharma: Some preamble here describing how PG organizes files here and where chunks fit in would make it easier to read/understand for someone new.

002-storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+- When database is smaller then a chunk size we naturally can store them in one chunk (since their page_key would fit in some chunk's [hi, lo) range).
```
adsharma: nit: *than* a chunk size

002-storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+<img width="937" alt="Screenshot_2021-02-22_at_16 49 17" src="https://user-images.githubusercontent.com/284219/108729836-ffcbd200-753b-11eb-9412-db802ec30021.png">
+
+Few databases are stored in one chunk, replicated three times
```
adsharma: Is this for simplicity? Do we want to consider availability zones? A failure of two machines would make the database unavailable.

002-storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+- When database can't fit into one storage node it can occupy lots of chunks that were split while database was growing. Chunk placement on nodes is controlled by us with some automatization, but we alway may manually move chunks around the cluster.
```
adsharma: I would promote the definition of keyspace below to be earlier in the document and then describe that we're using range partitioning on that key space.

002-storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+<img width="940" alt="Screenshot_2021-02-22_at_16 49 10" src="https://user-images.githubusercontent.com/284219/108729815-fb071e00-753b-11eb-86e0-be6703e47d82.png">
+
+Here one big database occupies two set of nodes. Also some chunks were moved around to restore replication factor after disk failure. In this case we also have "sharded" storage for a big database and issue wal writes to different chunks in parallel.
+
+## **Chunk placement strategies**
+
+There are few scenarios where we may want to move chunks around the cluster:
```
adsharma: Do we want to add load balancing?

002-storage.md
```diff
@@ -0,0 +1,186 @@
...
+(pg_id, db_id, ... , pageno=42, lsn=250) (per-page diff)
+(pg_id, db_id, ... , pageno=42, lsn=270) (per-page diff)
+(pg_id, db_id, ... , pageno=5000, lsn=100) (full 8k page data)
+```
+
+So query for `pageno=42 up to lsn=260` would need to find closest entry less then this key, iterate back to the latest full page and iterate forward to apply diffs. How often page is materialized in lsn-version sequence is up to us -- let's say each 5th version should be a full page.
+
+### **Page deletion**
+
+To delete old pages we insert blind deletion marker `(pg_id, db_id, #trim_lsn < 150)` into a lsm tree. During merges such marker would indicate that all pages with smaller lsn should be discarded. Delete marker will travel down the tree levels hierarchy until it reaches last level. In non-PITR scenario where old page version are not needed at all such deletion marker would (in average) prevent old page versions propagation down the tree -- so all bloat would concentrate at higher tree layers without affecting bigger bottom layers.
```
adsharma, line 72: LSMs such as rocksdb have their own internal LSN/seqno. We should look into whether we can exercise enough control over them to get what we want. If we can, this operation will be as simple as delete(key) @ LSM's lsn which is `somefunc(150)`.

adsharma (commented):
This is a very good start. Some suggestions to reorganize content with links so someone not super familiar with PG internals can grok what's going on.

002-storage.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+## **Chunk**
+
+Chunk is responsible for storing pages potentially from different databases and relations. Each page is addressed by a lexicographically ordered tuple (****page_key****) with following fields:
+
+- `pg_id` -- unique id of given postgres instance (or postgres cluster as it is called in postgres docs)
+- `db_id` -- database that was created by 'CREATE DATABASE' in a given postgres instance
+- `db_timeline` -- used to create Copy-on-Write instances from snapshots, described later
+- `rel_id` -- tuple of (relation_id, 0) for tables and (indexed_relation_id, rel_id) for indices. Done this way so table indices were closer to table itself on our global key space.
```
adsharma, line 46: This is a very good idea! Did you mean `(rel_id, indexed_relation_id)` here?

A related question - if we have a large relation (say 10s of TB), how do we ensure better locality of the index pages and heap pages?

Presumably someone will stackoverflow/websearch this problem and discover table partitioning. If they do, do we have any control over how the child rel_ids are allocated? Can we try to arrange for them to be sequential in this key space?

adsharma: Perhaps we are saying the same thing here. `(table_relation_id, index_relation_id)` would be less ambiguous.

knizhnik:
> 
> Yeah, the WAL is certainly the most straightforward format. Let's start with that.
> 
> In the future, we can optimize: instead of sending the whole WAL, the read-only node can "subscribe" to receive only WAL that's applicable to the pages that the node has in cache. Or, only send an invalidation stream, not containing the actual changes, but just the page # of modified pages. Or, if a replica is jumping from LSN X to (LSN X+ 1000), don't stream all the individual WAL records in that range, but a combined "diff" between X and X+1000.
> 
> But let's start with just the raw WAL stream.

Ok, looks like I have reinvented the bicycle:(
After reading more documentation I realized that both Amazon and Socrates are doing the same: streaming WAL from writer to replica.

But I still do not understand whether we need to have direct connection (replication stream) from master to replica?
It seems to be the most obvious and simplest solution. But how in this case we will handle crash of master  and provide execution of long queries at replicas without retarding vacuum at master?

As an alternative we can stream changes to replica from storage nodes. In this case we can maintain some artificial gap in applying WAL at replica. Just maintain xmin->LSN cache and adjust LSN according to the oldest xmin.

But it leads to the new question: should we scatter WAL records or broadcast WAL page as it is. When applying page, storage node may just skip records which update "foreign" chunks. In this case we will send more data through network, but can use broadcast mechanism. Also in this case and storage node can stream data to replica. Otherwise we should somehow reconstruct scattered WAL pages before sending them to replica and storing in S3. But we are loosing parallel WAL writes  optimization (yes, still all N storage nodes can flush WAL in parallel, but all of them will flush the same amount of data as produced by master, instead of 1/N).

So if among this three components:
- sending data through network
- flushing data to non-volatile media
- applying transactions
flushing is most expensive, scattering WAL may really be useful.

Socrates  propose very good principle: durability should be separated from availability.
For durability we need to write WAL at quorum of N storage nodes (where N is our redundancy level).
For availability we may need to spawn more then N storage nodes. More storage nodes may be needed to cache chunks of huge databases (or large number of small databases). What is not clear to me how WAL fits into this picture. Unlike data segments, computation nodes most frequently need most recent WAL segments. But original, non-scattered segments.

adsharma:
> But it leads to the new question: should we scatter WAL records or broadcast WAL page as it is.

I don't think it's an either-or question. We might have to do both. Scatter the WAL so we don't pay the network cost (especially on the master, where it could become a bottleneck) and broadcast it among the N nodes that host a particular chunk.

> without retarding vacuum at master?

You mean without slowing down vacuum at the master? The r-word has been in the news lately for all the wrong reasons, so let's try an alternative.

knizhnik:
> Scatter the WAL so we don't pay the network cost (especially on the master, where it could become a bottleneck) and broadcast it among the N nodes that host a particular chunk.
from my point of view, it should not affect master at all. At master we should have just one wal sender which normally stream WAL to some proxy. And all further broadcasting/scattering/write quorum logic should be implemented in this proxy. 

The question is whether this proxy need to parser WAL records and determine which blocks are affected and then compose new WAL chunks to be sent to the correspondent storage nodes. Or it should just broadcast WAL pages as it is to all nodes. I do no think that parsing itself will add noticeable overhead. But the problem is how we will then gather WAL records from different chunks to send to replicas and to S3.

> You mean without slowing down vacuum at the master?
Yes, if using hot_stadby_feedback we protect old version at master from vacuum, thren it may have negative impact on performance. I am not familiar with real Postgres cluster support in production, so I am sure how critical this problem is.

002-storage.md
kelvich: Yeah, that's in todo.

002-storage.md
kelvich: Each chunk is stored on N nodes, so failure of one would not affect availability. I was suggesting to store different databases in one chunks in order to simplify management of case with many small databases.

002-storage.md
```diff
@@ -0,0 +1,186 @@
...
+
+To delete old pages we insert blind deletion marker `(pg_id, db_id, #trim_lsn < 150)` into a lsm tree. During merges such marker would indicate that all pages with smaller lsn should be discarded. Delete marker will travel down the tree levels hierarchy until it reaches last level. In non-PITR scenario where old page version are not needed at all such deletion marker would (in average) prevent old page versions propagation down the tree -- so all bloat would concentrate at higher tree layers without affecting bigger bottom layers.
+
+### **Recovery**
+
+Upon storage node restart recent WAL files are applied to appropriate pages and resulting pages stored in lsm memtable. So this should be fast since we are not writing anything to disk.
+
+### **Checkpointing**
+
+No such mechanism is needed. Or we may look at the storage node as at kind of continuous chekpointer.
```
funny-falcon, line 80: afaik, lsm still needs some kind of "checkpoint". At least leveldb and its descendants have metadata file. Doubtfully metadata file could be updated continuously.

PR #2, opened 2021-02-27
003 laptop cli.md

kelvich:
[Rendered](https://github.com/libzenith/rfcs/blob/003-laptop-cli.md/003-laptop-cli.md)

### DIFF
--- a/003-laptop-cli.md
+++ b/003-laptop-cli.md
@@ -2,17 +2,112 @@
 
 Zenith CLI as it is described here mostly resides on the same conceptual level as pg_ctl/initdb/pg_recvxlog/etc and replaces some of them in an opinionated way. I would also suggest bundling our patched postgres inside zenith distribution at least at the start.
 
-The most important concept here is a snapshot, which can be created/moved/sent/exported. Also, we may start temporary read only postgres instance over any local snapshot. A more complex scenarios would consist of several basic operations over snapshots.
+This proposal is focused on managing local installations. For cluster operations, different tooling would be needed. The point of integration between the two is storage URL: no matter how complex cluster setup is it may provide an endpoint where the user may push snapshots.
+
+The most important concept here is a snapshot, which can be created/pushed/pulled/exported. Also, we may start temporary read-only postgres instance over any local snapshot. A more complex scenario would consist of several basic operations over snapshots.
+
+# Possible usage scenarios
+
+## Install zenith, run a postgres
+
+```
+> brew install pg-zenith 
+> zenith pg create # creates pgdata with default pattern pgdata$i
+> zenith pg list
+ID            PGDATA        USED    STORAGE            ENDPOINT
+primary1      pgdata1       0G      zenith-local       localhost:5432
+>
+```
+
+## Import standalone postgres to zenith
+
+```
+> zenith snapshot import --from=basebackup://replication@localhost:5432/ oldpg
+[====================------------] 60% | 20MB/s
+> zenith snapshot list
+ID          SIZE        PARENT
+oldpg       5G          -
+
+> zenith pg create --snapshot oldpg
+Started postgres on localhost:5432
+
+> zenith pg list
+ID            PGDATA        USED    STORAGE            ENDPOINT
+primary1      pgdata1       5G      zenith-local       localhost:5432
+
+> zenith snapshot destroy oldpg
+Ok
+> 
+```
+
+Also, we may start snapshot import implicitly by looking at snapshot schema
+
+```
+> zenith pg create --snapshot basebackup://replication@localhost:5432/
+Downloading snapshot... Done.
+Started postgres on localhost:5432
+Destroying snapshot... Done.
+```
+
+## Pull snapshot with some publicly shared database
+
+Since we may export the whole snapshot as one big file (tar of basebackup, maybe with some manifest) it may be shared over conventional means: http, ssh, [git+lfs](https://docs.github.com/en/github/managing-large-files/about-git-large-file-storage).
+
+```
+> zenith pg create --snapshot http://learn-postgres.com/movies_db.zenith movies
+```
+
+## Rollback database to the snapshot
+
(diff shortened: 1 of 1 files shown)


003-laptop-cli.md
```diff
@@ -0,0 +1,146 @@
+# Command line interface (end-user)
+
+Zenith CLI as it is described here mostly resides on the same conceptual level as pg_ctl/initdb/pg_recvxlog/etc and replaces some of them in an opionated way. I would also suggest to bundle our patched postgres inside zenith distribution at least on the start.
+
+
+## storage
+
+Storage is either page service or s3. User may create a database in pagestore and create/move *snapshots* and *pirt regions* in both pagestore and s3. Storage is a concept similar to `git remote`. After installation I imagine two storages are available by default: local pagestore and zenith cloud pagestore.
+
+XXX: is there any better ideas on how to call pagestore? may be just zenith? zenith-smgr?
```
adsharma: I like `zstore`

adsharma (commented):
Thanks for writing this up!

High level:

* We can look at other CLIs such as ceph
** https://docs.ceph.com/en/latest/rados/operations/add-or-rm-osds/
** https://gist.github.com/pertoft/a4250bd2826a1be5f341c71c30e9e7e5

* Do you want to add a placeholder for zenith push, zenith log?
* Do we need the ability to show all the protection groups?
* Control replication factor per protection group?

* How do read replicas work?
* Do we need the concept of a volume?
* How do PG concepts like tablespace, database, schema map to zenith?
* How do we expose concepts like LSN, volume complete, protection group complete, segment complete?
* Repair

kelvich:
>     * We can look at other CLIs such as ceph
>       ** https://docs.ceph.com/en/latest/rados/operations/add-or-rm-osds/
>       ** https://gist.github.com/pertoft/a4250bd2826a1be5f341c71c30e9e7e5

I was mostly looking at zfs/zpool CLI and git.

>     * Do you want to add a placeholder for zenith push, zenith log?

```
zenith snapshot push
zenith snapshot list
```

>     * Do we need the ability to show all the protection groups?
> 
>     * Control replication factor per protection group?
> 
>     * How do read replicas work?
> 
>     * Do we need the concept of a volume?
> 
>     * How do PG concepts like tablespace, database, schema map to zenith?
> 
>     * How do we expose concepts like LSN, volume complete, protection group complete, segment complete?
> 
>     * Repair

I've added disclaimer to the top about scope of this CLI. That one is mostly for managing laptop/small installation and ability to share/exchange snapshots. Things that you are talking about are important but I imagine that it would be a different tooling there.

adsharma:
https://litestream.io/reference/snapshots/ - does something similar. Worth
looking at?

003-laptop-cli.md
```diff
@@ -0,0 +1,262 @@
...
+
+
+
+## pg
+
+Manages postgres data directories and can start postgreses with proper configuration. An experienced user may avoid using that (except pg create) and configure/run postgres by themself.
+
+Pg is a term for a single postgres running on some data. I'm trying to avoid here separation of datadir management and postgres instance management -- both that concepts bundled here together.
+
+**zenith pg create** [--no-start --snapshot --cow] -s storage-name -n pgdata
```
lubennikovaav, line 143: What does '-n pgdata' mean here?

003-laptop-cli.md
```diff
@@ -0,0 +1,262 @@
...
+
+--no-start: just init datadir without creating 
+
+--snapshot snap: init from the snapshot. Snap is a name or URL (zenith.tech/stas/mystore/snap1)
+
+--cow: initialize Copy-on-Write data directory on top of some snapshot (makes sense if it is a snapshot of currently running a database)
+
+**zenith pg destroy**
+
+**zenith pg start** [--replica] pgdata
```
lubennikovaav, line 155: Does replica here mean 'read-only zenith node'  or something else?

003-laptop-cli.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+Resetting the database to some state in past would require creating a snapshot on some lsn / time in this pirt area.
+
+# Manual
+
+## storage
+
+Storage is either zenith pagestore or s3. Users may create a database in a pagestore and create/move *snapshots* and *pitr regions* in both pagestore and s3. Storage is a concept similar to `git remote`. After installation, I imagine one local storage is available by default.
+
```
lubennikovaav, line 112: I don't get this part. Shouldn't we always have zenith pagestore, either local or remote? And optional (?) s3 for wall archive and base images. 

Besides, in my understanding, zenith storage is not just a datadir - also includes page server and wal-safekeeper processes. How do we start/stop them? 

We can start them along with a master pg node. But what if we want to shut down master for a while and work with read-only nodes over the storage?

003-laptop-cli.md
kelvich: I meant that '-n' is the name of the database and '-s' is s3/pageserver/remote-pageserver. May be '-s' is unnecessary degree of freedom right now.

003-laptop-cli.md
kelvich: I meant starting actual replica for the same database.

003-laptop-cli.md
kelvich: > I don't get this part. Shouldn't we always have zenith pagestore, either local or remote? And optional (?) s3 for wall archive and base images.

Depending on the format of files in the page store we may have the same format on s3 too -- if it is stored in LSM on the pageserver we never update SSTs and may just push them down to s3. It still would be some images + wal records just with more involved format.

Lets start here with the pageserver now, either local or remote.

> Besides, in my understanding, zenith storage is not just a datadir - also includes page server and wal-safekeeper processes. How do we start/stop them?

For a laptop installation I would just bundle them together for simplicity. I'm thinking of following setup:

* User install zenith via `brew install libzenith/tap/zenith` or `curl -L https://zenith.tech/install.sh | sh` for linux
* zenith start / stop is controlled by os-specific commands. `brew services start zenith` or `service start zenith`
* when zenith is started it is zenithd daemon that has safekeeper/pageserver/http-server inside (for starters we may have python sketch of zenithd who will just have http part and will run safekeeper/pageserver as child processes)
* zenith cli talks to zenithd http and asks to start/restore something

kelvich:
> https://litestream.io/reference/snapshots/ - does something similar. Worth looking at?
> […](#)

Yeah, looks cool! Thanks for bringing this up.

PR #3, opened 2021-03-02
RFC: Durability and consensus.

hlinnaka:
I spent the day getting up to speed on Paxos, Raft and Aurora papers. Here's my vision and understanding of how the consensus would work. @adsharma does this align with your thoughts?

All of this is could be implemented for unmodified PostgreSQL, as external processes that speak the streaming replication with PostgreSQL. That could become popular in the wider PostgreSQL ecosystem. Currently, if you want to set up a highly available PostgreSQL cluster, you use tools like Patroni, Zookeeper, Consul or pg_autofailover to decide who's the primary. And WAL-G for WAL archiving. But I don't think those tools give you a quorum-based synchronously replicated solution. A standalone project that implements these WAL safekeepers and the consensus among them, without having to rely on external tools, would be nice.

- Heikki

### DIFF
--- a/Durability.md
+++ b/Durability.md
@@ -1,0 +1,218 @@
+Durability & Consensus
+======================
+
+When a transaction commits, a commit record is generated in the WAL.
+When do we consider the WAL record as durable, so that we can
+acknowledge the commit to the client and be reasonably certain that we
+will not lose the transaction?
+
+Zenith uses a group of WAL safekeeper nodes to hold the generated WAL.
+A WAL record is considered durable, when it has been written to a
+majority of WAL safekeeper nodes. In this document, I use 5
+safekeepers, because I have five fingers. A WAL record is durable,
+when at least 3 safekeepers have written it to disk.
+
+First, assume that only one primary node can be running at a
+time. This can be achieved by Kubernetes or etcd or some
+cloud-provider specific facility, or we can implement it
+ourselves. These options are discussed in later chapters.  For now,
+assume that there is a Magic STONITH Fairy that ensures that.
+
+In addition to the WAL safekeeper nodes, the WAL is archived in
+S3. WAL that has been archived to S3 can be removed from the
+safekeepers, so the safekeepers don't need a lot of disk space.
+
+
+                                +----------------+
+                        +-----> | WAL safekeeper |
+                        |       +----------------+
+                        |       +----------------+
+                        +-----> | WAL safekeeper |
++------------+          |       +----------------+
+|  Primary   |          |       +----------------+
+| Processing | ---------+-----> | WAL safekeeper |
+|   Node     |          |       +----------------+
++------------+          |       +----------------+
+            \           +-----> | WAL safekeeper |
+             \          |       +----------------+
+              \         |       +----------------+
+               \        +-----> | WAL safekeeper |
+                \               +----------------+
+                 \
+                  \
+                   \
+                    \
+                     \      +--------+
+					  \		|        |
+					   +-->	|   S3   |
+							|        |
+                            +--------+
+
+
+Every WAL safekeeper holds a section of WAL, and a VCL value.
+The WAL can be divided into three portions:
+
+
+                                    VCL                   LSN
+                                     |                     |
+                                     V                     V
+.................ccccccccccccccccccccXXXXXXXXXXXXXXXXXXXXXXX
+Archived WAL       Completed WAL          In-flight WAL
+
+
+Note that all this WAL kept in a safekeeper is a contiguous section.
+This is different from Aurora: In Aurora, there can be holes in the
+WAL, and there is a Gossip protocol to fill the holes. That could be
+implemented in the future, but let's keep it simple for now. WAL needs
+to be written to a safekeeper in order. However, during crash
+recovery, In-flight WAL that has already been stored in a safekeeper
+can be truncated or overwritten.
+
+The Archived WAL has already been stored in S3, and can be removed from
+the safekeeper.
+
+The Completed WAL has been written to at least three safekeepers. The
+algorithm ensures that it is not lost, when at most two nodes fail at
+the same time.
+
+The In-flight WAL has been persisted in the safekeeper, but if a crash
+happens, it may still be overwritten or truncated.
+
+
+The VCL point is determined in the Primary. It is not strictly
+necessary to store it in the safekeepers, but it allows some
+optimizations and sanity checks and is probably generally useful for
+the system as whole. The VCL values stored in the safekeepers can lag
+behind the VCL computed by the primary.
+
+
+Primary node Normal operation
+-----------------------------
+
+1. Generate some WAL.
+
+2. Send the WAL to all the safekeepers that you can reach.
+
+3. As soon as a quorum of safekeepers have acknowledged that they have
+   received and durably stored the WAL up to that LSN, update local VCL
+   value in memory, and acknowledge commits to the clients.
+
+4. Send the new VCL to all the safekeepers that were part of the quorum.
+   (Optional)
+
+
+Primary Crash recovery
+----------------------
+
+When a new Primary node starts up, before it can generate any new WAL
+it needs to contact a majority of the WAL safekeepers to compute the
+VCL. Remember that there is a Magic STONITH fairy that ensures that
+only node process can be doing this at a time.
+
+1. Contact all WAL safekeepers. Find the Max((Epoch, LSN)) tuple among the ones you
+   can reach. This is the Winner safekeeper, and its LSN becomes the new VCL.
+
+2. Update the other safekeepers you can reach, by copying all the WAL
+   from the Winner, starting from each safekeeper's old VCL point. Any old
+   In-Flight WAL from previous Epoch is truncated away.
+
+3. Increment Epoch, and send the new Epoch to the quorum of
+   safekeepers.  (This ensures that if any of the safekeepers that we
+   could not reach later come back online, they will be considered as
+   older than this in any future recovery)
+
+You can now start generating new WAL, starting from the newly-computed
+VCL.
+
+Optimizations
+-------------
+
+As described, the Primary node sends all the WAL to all the WAL safekeepers. That
+can be a lot of network traffic. Instead of sending the WAL directly from Primary,
+some safekeepers can be daisy-chained off other safekeepers, or there can be a
+broadcast mechanism among them. There should still be a direct connection from the
+each safekeeper to the Primary for the acknowledgments though.
+
+Similarly, the responsibility for archiving WAL to S3 can be delegated to one of
+the safekeepers, to reduce the load on the primary.
+
+
+Magic STONITH fairy
+-------------------
+
+Now that we have a system that works as long as only one primary node is running at a time, how
+do we ensure that?
+
+1. Use etcd to grant a lease on a key. The primary node is only allowed to operate as primary
+   when it's holding a valid lease. If the primary node dies, the lease expires after a timeout
+   period, and a new node is allowed to become the primary.
+
+2. Use S3 to store the lease. S3's consistency guarantees are more lenient, so in theory you
+   cannot do this safely. In practice, it would probably be OK if you make the lease times and
+   timeouts long enough. This has the advantage that we don't need to introduce a new
+   component to the architecture.
+
+3. Use Raft or Paxos, with the WAL safekeepers acting as the Acceptors to form the quorum. The
+   next chapter describes this option.
+
+
+Built-in Paxos
+--------------
+
+The WAL safekeepers act as PAXOS Acceptors, and the Processing nodes
+as both Proposers and Learners.
+
+Each WAL safekeeper holds an Epoch value in addition to the VCL and
+the WAL. Each request by the primary to safekeep WAL is accompanied by
+an Epoch value. If a safekeeper receives a request with Epoch that
+doesn't match its current Accepted Epoch, it must ignore (NACK) it.
+(In different Paxos papers, Epochs are called "terms" or "round
+numbers")
+
+When a node wants to become the primary, it generates a new Epoch
+value that is higher than any previously observed Epoch value, and
+globally unique.
+
+
+Accepted Epoch: 555                VCL                   LSN
+                                     |                     |
+                                     V                     V
+.................ccccccccccccccccccccXXXXXXXXXXXXXXXXXXXXXXX
+Archived WAL       Completed WAL          In-flight WAL
+
+
+Primary node startup:
+
+1. Contact all WAL safekeepers that you can reach (if you cannot
+   connect to a quorum of them, you can give up immediately). Find the
+   latest Epoch among them.
+
+2. Generate a new globally unique Epoch, greater than the latest Epoch
+   found in previous step.
+
+2. Send the new Epoch in a Prepare message to a quorum of
+   safekeepers. (PAXOS Prepare message)
+
+3. Each safekeeper responds with a Promise. If a safekeeper has
+   already made a promise with a higher Epoch, it doesn't respond (or
+   responds with a NACK). After making a promise, the safekeeper stops
+   responding to any write requests with earlier Epoch.
+
+4. Once you have received a majority of promises, you know that the
+   VCL cannot advance on the old Epoch anymore. This effectively kills
+   any old primary server.
+
+5. Find the highest written LSN among the quorum of safekeepers (these
+   can be included in the Promise messages already). This is the new
+   VCL.  If a new node starts the election process after this point,
+   it will compute the same or higher VCL.
+
+6. Copy the WAL from the safekeeper with the highest LSN to the other
+   safekeepers in the quorum, using the new Epoch. (PAXOS Accept
+   phase)
+
+7. You can now start generating new WAL starting from the VCL. If
+   another process starts the election process after this point and
+   gains control of a majority of the safekeepers, we will no longer
+   be able to advance the VCL.
+


Durability.md
```diff
@@ -0,0 +1,218 @@
...
+
+3. As soon as a quorum of safekeepers have acknowledged that they have
+   received and durably stored the WAL up to that LSN, update local VCL
+   value in memory, and acknowledge commits to the clients.
+
+4. Send the new VCL to all the safekeepers that were part of the quorum.
+   (Optional)
+
+
+Primary Crash recovery
```
adsharma, line 104: The hard part is:

* Primary failover to a secondary
* Operation under network partition

https://www.sougou.io/distributed-durability-in-mysql/

adsharma (commented):
This sounds equivalent to writing WAL to a filesystem backed by distributed consensus.

These options will give us a durable WAL that all participants agree on, but do not provide tx isolation guarantees of a single machine postgres.

I have a bunch of ideas I'm in the process of writing up. Will share once complete.

Durability.md
```diff
@@ -0,0 +1,218 @@
...
+   any old primary server.
+
+5. Find the highest written LSN among the quorum of safekeepers (these
+   can be included in the Promise messages already). This is the new
+   VCL.  If a new node starts the election process after this point,
+   it will compute the same or higher VCL.
+
+6. Copy the WAL from the safekeeper with the highest LSN to the other
+   safekeepers in the quorum, using the new Epoch. (PAXOS Accept
+   phase)
```
kelvich, line 212: 👍 yeah, I also was thinking about that place. Raft does this differently in a way that we can't support since processing nodes are not proposing themselves. So following paxos here is better.

hlinnaka:
>These options will give us a durable WAL that all participants agree on, but do not provide tx isolation guarantees of a single machine postgres.

Right, TX isolation is a completely separate, orthogonal, problem.

Durability.md
```diff
@@ -0,0 +1,218 @@
...
+                                    VCL                   LSN
+                                     |                     |
+                                     V                     V
+.................ccccccccccccccccccccXXXXXXXXXXXXXXXXXXXXXXX
+Archived WAL       Completed WAL          In-flight WAL
+
+
+Note that all this WAL kept in a safekeeper is a contiguous section.
+This is different from Aurora: In Aurora, there can be holes in the
+WAL, and there is a Gossip protocol to fill the holes. That could be
```
kelvich, line 65: I was imagining that all that wal-with-holes thing is to make safekeeper recovery easier. If one safekeeper reboots it will need to get missing WAL from somewhere. Postgres now can hold WAL for absent safekeeper, but that can impact processing node (disk usage & disk space). Other option is not to hold WAL at processing node at all and start receiving new WAL immediately after reboot. When safekeeper will have first new WAL it can determine gap between its own WAL and incoming WAL and ask neighbors to provide corresponding pages/WAL.

Durability.md
arssher: > in a way that we can't support since processing nodes are not proposing themselves

But why? It would be fine to allow VCL advancement by the new primary only after old missing WAL (preserving the epoch) copied on majority + a (dummy) entry from the new epoch is also written on it, like Raft does. This choice is argued in 5.4.2 section of
https://raft.github.io/raft.pdf

Restamping records with the new epoch instead (proposed here) is possible, but sounds more complicated and expensive; note that you must renumber them *everywhere*, including all servers who already have the record.

hlinnaka: Hmm. I don't think we need to stamp the records with epoch at all. Each WAL safekeeper has a *current* epoch value, and the WAL. The current epoch is just a variable, you don't need to write it in the log, or associate it with any particular WAL records.

hlinnaka: I think what I'm proposing is Multi-Paxos: https://en.wikipedia.org/wiki/Paxos_(computer_science)#Multi-Paxos. Mapping the terms:

The Processing node is a Client, Proposer and Learner. The WAL safekeepers are Acceptors.

When the processing node starts up and wants to become the primary, it initiates the elections, and perform the full Paxos protocol. This is the "Multi-Paxos without failures" sequence. Subsequent writes to the WAL can skip the phase 1, following the "Multi-Paxos when phase 1 can be skipped" sequence.

kelvich: > This choice is argued in 5.4.2 section of

Yeah, I was talking about that election restriction in 5.4.1: "Raft uses the voting process to prevent a candidate from winning an election unless its log contains all committed entries". If we assume that processing node is stateless it have no logs but we allow it to be elected and there is nothing to replicate in previous term. So we need to brew something different: either some recovery mechanism through master or allow safe keepers to communicate between each other.

> I don't think we need to stamp the records with epoch at all.

Are you talking about network or disk? In network all records should be packed within some container stamped with epoch so that safe keeper can distinguish between records from competing masters. With disk I can't came up right away with example when we need to know term for each record -- I'll think more.

hlinnaka:
If we used Raft for this, how would the Raft concepts map to ours? The WAL safekeepers would be Raft "Servers", right? And the processing nodes would be Clients? What are the "Commands" that the processing nodes send? Each "command" a piece of WAL to append to the log?

Yeah, I think that would work too. We need to ensure that there is only one node generating the WAL, so the WAL safekeeper that acts as the leader would need to ensure that only one primary node is connected at a time, with a heartbeat or something.

I find the Paxos algorithm easier to map to our concepts, with its separation of the Proposer and Acceptor roles. And I like the fact that there doesn't need to be any direct communication between the Acceptors (although in practice we would probably want the WAL safekeepers to forward the WAL to each other, to reduce the processing node's load). But I'm not sure which one would be simpler to implement in practice.

Either way, it can be done, and the choice between Raft or Paxos doesn't affect the rest of the architecture. That's enough to start implementing.

Durability.md
hlinnaka: > > This choice is argued in 5.4.2 section of
> 
> Yeah, I was talking about that election restriction in 5.4.1: "Raft uses the voting process to prevent a candidate from winning an election unless its log contains all committed entries". If we assume that processing node is stateless it have no logs but we allow it to be elected and there is nothing to replicate in previous term. So we need to brew something different: either some recovery mechanism through master or allow safe keepers to communicate between each other.

I'm imagining that the processing node is the Client, not a Server. The primary processing node connects to the leader WAL safekeeper. Yeah, the safekeepers will need to communicate with each other in Raft.

> > I don't think we need to stamp the records with epoch at all.
> 
> Are you talking about network or disk? In network all records should be packed within some container stamped with epoch so that safe keeper can distinguish between records from competing masters. With disk I can't came up right away with example when we need to know term for each record -- I'll think more.

On disk. You're right that you need the epoch in the network requests. Or you rely on the leader safekeeper to only allow one primary processing node to be connected at a time, and when you connect you first ask the safekeeper what the current end-of-WAL is, before you start to append.

kelvich: > I'm imagining that the processing node is the Client, not a Server. 

Wouldn't that lead to same race between two competing clients?

I was imagining processing node being a Server which can propose himself, but can't vote (so destined to be leader) and safe keeper being a Server which can vote but can't propose (forever follower). Also in this setup we don't know amount of participants from the start: set of safekeepers is fixed but set of processing is potentially unbounded.
So more or less same setup as you proposed with Paxos.

adsharma:
Raft and paxos are 90% similar

https://arxiv.org/pdf/2004.05074.pdf
Differences called out in the conclusion section of the paper.

Also Fig 2 in this paper calls out the number of lines of code different
between
Raft and Paxos in blue. It starts from paxos and shows how raft is
different.
https://dl.acm.org/doi/pdf/10.1145/3293611.3331595

Heidi has a similar image posted somewhere on twitter where she annotates
the 1 page raft algorithm in red.

Given the number of productionized raft instances and paxos instances in
the last 5 years, I think we should start with raft as the starting point,
since it's easier to debug. I could be convinced otherwise if you're
proposing some variant of paxos that has a significant demonstrated benefit.

adsharma:
The Heidi highlighted image is also viewable in this video at this
timestamp:

https://www.youtube.com/watch?v=0K6kt39wyH0&t=3468s

adsharma:
https://twitter.com/heidiann360/status/1254701717150720000

adsharma:
> If we used Raft for this, how would the Raft concepts map to ours? 

Described in the doc titled "Consensus for Disaggregated Storage"
WAL = raft log. We don't need to persist them to disk twice.
Compute node = leader or follower
WAL safekeeper node = learner (can't vote itself to be a primary)

Ditto for page store.

It's possible that we model them as one quorum and add additional rules to describe capabilities of each type of node. I think it's harder to prove correctness.

Hence my attempt to have multiple quorums and then add rules between quorums.

Durability.md
```diff
@@ -0,0 +1,218 @@
...
+-------------------
+
+Now that we have a system that works as long as only one primary node is running at a time, how
+do we ensure that?
+
+1. Use etcd to grant a lease on a key. The primary node is only allowed to operate as primary
+   when it's holding a valid lease. If the primary node dies, the lease expires after a timeout
+   period, and a new node is allowed to become the primary.
+
+2. Use S3 to store the lease. S3's consistency guarantees are more lenient, so in theory you
```
funny-falcon, line 150: This year Amazon claims they've made S3 operations strongly consistent.
https://aws.amazon.com/ru/s3/consistency/
https://aws.amazon.com/ru/blogs/aws/amazon-s3-update-strong-read-after-write-consistency/

Durability.md
```diff
@@ -0,0 +1,218 @@
...
+   responds with a NACK). After making a promise, the safekeeper stops
+   responding to any write requests with earlier Epoch.
+
+4. Once you have received a majority of promises, you know that the
+   VCL cannot advance on the old Epoch anymore. This effectively kills
+   any old primary server.
+
+5. Find the highest written LSN among the quorum of safekeepers (these
+   can be included in the Promise messages already). This is the new
+   VCL.  If a new node starts the election process after this point,
```
funny-falcon, line 207: Nope. Only after first record of new Epoch will be commited to majority. ~Read Raft paper carefully.~ At least if you follow Raft and don't provide some additional guarantees.

Durability.md
```diff
@@ -0,0 +1,218 @@
...
+2. Send the new Epoch in a Prepare message to a quorum of
+   safekeepers. (PAXOS Prepare message)
+
+3. Each safekeeper responds with a Promise. If a safekeeper has
+   already made a promise with a higher Epoch, it doesn't respond (or
+   responds with a NACK). After making a promise, the safekeeper stops
+   responding to any write requests with earlier Epoch.
+
+4. Once you have received a majority of promises, you know that the
+   VCL cannot advance on the old Epoch anymore. This effectively kills
```
funny-falcon, line 202: Nope. Reellection could happen and old primary could replicate more records of old epoch.

hlinnaka: After a safekeeper has made a Promise on the new Epoch, it will no longer accept records with the old Epoch. After a majority of safekeepers have made a Promise on the new Epoch, the old primary with old Epoch can no longer get a majority on any changes on the old Epoch, so it can no longer advance the VCL.

If I'm missing something, can you give a step-by-step example of a sequence where that could happen?

Durability.md
hlinnaka: This is based on Paxos, which is slightly different from Raft on that point, if I've understood them correctly. Can you give a step-by-step sequence of events of where that would happen?

Durability.md
funny-falcon: For simplicity lets have 3 safekeepers A, B, C
Primary P1 writes records for Epoch E1: R1_1, R1_2, R1_3 (in format of `R<EPOCH>_<LSN>`)
R1_1 replicated to all A, B, C
R1_2 replicated to B
R1_3 didn't replicated
Primary had got a pause
New primary P2 has promises for epoch E2 from A and C. It designated VCL as R1_1. And it even replicates R2_2 to A. Then it also stalls
Old primary P1 awaken and has promises for epoch E3 from B and C. Now it replicates R1_2 to C and R1_3 to B and C - yep VCL were advanced for epoch E1.

funny-falcon: Even if considered step 6 (copy WAL from safekeeper with highest LSN), then still VCL will be advanced from R1_1 to R1_2.

Durability.md
funny-falcon: For simplicity lets have 3 safekeepers A, B, C
Primary P1 writes records for Epoch E1: R1_1, R1_2 (in format of `R<EPOCH>_<LSN>`)
R1_1 replicated to all A, B, C
R1_2 replicated to B
Primary had got a pause
New primary P2 has promises for epoch E2 from A and B. It designated VCL as R1_2. It copies R1_2 to A and replicates R2_3 to A. Then it failed.
New primary P3 has promises for epoch E3 from A and B. It designates VCL as R2_3. And it failed immediately.
New primary P4 has promises for epoch E4 from B and C. It designates VCL as R1_2.

funny-falcon: Therefore, VCL have to be defined not as "highest written LSN", but as highest "common LSN". And it could advance only after step 6, when WAL is copied from safekeeper with the highest LSN to quorum.

hlinnaka: Gotcha, thanks! Right, you must finish step 6 before you can consider the "highest LSN" as the durable VCL.

Durability.md
hlinnaka: > For simplicity lets have 3 safekeepers A, B, C
> Primary P1 writes records for Epoch E1: R1_1, R1_2, R1_3 (in format of `R<EPOCH>_<LSN>`)
> R1_1 replicated to all A, B, C
> R1_2 replicated to B
> R1_3 didn't replicated
> Primary had got a pause
> New primary P2 has promises for epoch E2 from A and C. It designated VCL as R1_1. And it even replicates R2_2 to A. Then it also stalls

I'm with you so far.

> Old primary P1 awaken and has promises for epoch E3 from B and C. Now it replicates R1_2 to C and R1_3 to B and C - yep VCL were advanced for epoch E1.

Ah, here's where we differ in terminology, I think. The old primary cannot replicate R1_2 to C; C won't accept it because it already made a promise on E3. Instead, the old primary would replicate R3_2 to C, and R3_3 to B and C. In other words, it would replicate the same piece of WAL with the new Epoch. Does that make sense?

funny-falcon: > Ah, here's where we differ in terminology, I think. The old primary cannot replicate R1_2 to C; C won't accept it because it already made a promise on E3. Instead, the old primary would replicate R3_2 to C, and R3_3 to B and C. In other words, it would replicate the same piece of WAL with the new Epoch. Does that make sense?

Lets rephrase same thing without "Old Primary".

For simplicity lets have 3 safekeepers A, B, C
Primary P1 writes records for Epoch E1: R1_1(x1), R1_2(x2) (in format of `R<EPOCH>_<LSN>(value)` )
R1_1(x1) replicated to all A, B, C
R1_2(x2) replicated to B
Primary P1 died.

New primary P2 has promises for epoch E2 from A and C. It designated VCL as R1_1. And it wrotes R2_2(y1) to A. Then it also died.

Next primary P3 has got promises for epoch E3 from B and C.
It sees: B has records R1_1(x1) and R1_2(x2), C has record R1_1(x1).
It has to replicate R1_2(x2) to C due to step 6 in this algo.

Which way it will do it? Raft paper says it will be same R1_2(x2).
Do you mean it will be R3_2(x2)?

funny-falcon: Other example:
P1 performs R1_1(x1), R1_2(x2), R1_3(x3).
A got all R1_1(x1), R1_2(x2) and R1_3(x3).
B got R1_1(x1)
C also got R1_1(x1).
P1 died.

P2 has votes for E2 from B and C. P2 performs R2_2(y1) and wrotes it to B and C. Then it died.

Now we have:
A has R1_1(x1), R1_2(x2) and R1_3(x3)
B has R1_1(x1) and R2_2(y1)
C has R1_1(x1) and R2_2(y1)

Primary P3 has votes for E3 from A and B. Which records it will replicated in which side?

funny-falcon: Other possible (really) situation:
P1 performs R1_1(x1), R1_2(x2).
A got all R1_1(x1), R1_2(x2).
B got R1_1(x1)
C also got R1_1(x1).
P1 died.

P2 has votes for E2 from B and C. P2 performs R2_2(y1) and wrotes it to B and C. Then it died.

A has R1_1(x1), R1_2(x2)
B has R1_1(x1), R2_2(y1)
C has R1_1(x1).

P3 connects A and C for epoch E3. It replicates R1_2(x2) to C (either as R1_2(x2) or as R3_2(x2). And died.

A has R1_1(x1), R1_2(x2)
B has R1_1(x1), R2_2(y1)
C has R1_1(x1), R1_2(x2) (or R3_2(x2) if you like).

P4 connects to A and B for epoch E4. Which value it should replicate? x2 or y1?

Raft has an answer to this: R1_2(x2) is not considered as committed, because to consider it as committed P3 have to commit R3_3(z1) (ie write to both A and C).
If it does, then R1_2(x2) is committed, and P4 have to replicate R1_2(x2) and R3_3(z1) to B.
If it doesn't, then P4 have to replicate R2_2(y1) to A. Note, that it still will not be committed until P4 commits R4_3(thau1) as well.

funny-falcon: Raft is not separate algorithm, as one might think. Raft is well and thoroughly described MultiPaxos as it should be (ok, with some tuning: placing Proposer, Accepter and Learner at same processes). One have to understand quirks of Raft to clearly understand MultiPaxos.

hlinnaka: > > Ah, here's where we differ in terminology, I think. The old primary cannot replicate R1_2 to C; C won't accept it because it already made a promise on E3. Instead, the old primary would replicate R3_2 to C, and R3_3 to B and C. In other words, it would replicate the same piece of WAL with the new Epoch. Does that make sense?
> 
> Lets rephrase same thing without "Old Primary".
> 
> For simplicity lets have 3 safekeepers A, B, C
> Primary P1 writes records for Epoch E1: R1_1(x1), R1_2(x2) (in format of `R<EPOCH>_<LSN>(value)` )
> R1_1(x1) replicated to all A, B, C
> R1_2(x2) replicated to B
> Primary P1 died.
> 
> New primary P2 has promises for epoch E2 from A and C. It designated VCL as R1_1. And it wrotes R2_2(y1) to A. Then it also died.
> 
> Next primary P3 has got promises for epoch E3 from B and C.
> It sees: B has records R1_1(x1) and R1_2(x2), C has record R1_1(x1).
> It has to replicate R1_2(x2) to C due to step 6 in this algo.
> 
> Which way it will do it? Raft paper says it will be same R1_2(x2).
> Do you mean it will be R3_2(x2)?

That's right. Kind of. In this scheme, the Epoch number is just a single variable in the safekeeper. The WAL records are not "stamped" with the epoch, the safekeeper only stores its "current" epoch (see also discussion with @arssher above). I should make that more clear in the doc. Rephrasing that scenario with that in mind:

Primary P1 writes records for Epoch E1: R1(x1), R1(x2) (in format of `R<LSN>(value)` ). R1(x1) is replicated to all A, B, C. R2(x2) replicated to only B. Situation now:

A (E1): R1(x1)
B (E1): R1(x1), R2(x2)
C (E1): R1(x1)

Primary P1 dies.

New primary P2 gets promises for epoch E2 from A and C. It designates VCL as R1. And it writes R2(y1) to A. Then it also dies. Situation now:

A (E2): R1(x1), R2(y1)
B (E1): R1(x1), R2(x2)
C (E2): R1(x1)

Next primary P3 gets promises for epoch E3 from B and C.

A (E2): R1(x1), R2(y1)
B (E3): R1(x1), R2(x2)
C (E3): R1(x1)

P3 now replicates R2(x2) to C due to step 6 in this algo:

A (E2): R1(x1), R2(y1)
B (E3): R1(x1), R2(x2)
C (E3): R1(x1), R2(x2)

hlinnaka: > Other example:
> P1 performs R1_1(x1), R1_2(x2), R1_3(x3).
> A got all R1_1(x1), R1_2(x2) and R1_3(x3).
> B got R1_1(x1)
> C also got R1_1(x1).
> P1 died.
> 
> P2 has votes for E2 from B and C. P2 performs R2_2(y1) and wrotes it to B and C. Then it died.
> 
> Now we have:
> A has R1_1(x1), R1_2(x2) and R1_3(x3)
> B has R1_1(x1) and R2_2(y1)
> C has R1_1(x1) and R2_2(y1)
> 
> Primary P3 has votes for E3 from A and B. Which records it will replicated in which side?

Rephrasing this scenario too so that the safekeeper only stores the current Epoch:

P1 performs R1_1(x1), R1_2(x2), R1_3(x3). A got all R1_1(x1), R1_2(x2) and R1_3(x3). B got R1_1(x1) C also got R1_1(x1). P1 dies. Situation:

A (E1): R1(x1), R2(x2), R3(x3)
B (E1): R1(x1)
C (E1): R1(x1)

P2 has votes for E2 from B and C. P2 performs R2(y1) and writes it to B and C. Then it dies. Now we have:
A (E1): R1(x1), R2(x2), R3(x3)
B (E2): R1(x1), R2(y1)
C (E2): R1(x1), R2(y1)

Primary P3 gets Promises for E3 from A and B:
A (E3): R1(x1), R2(x2), R3(x3)
B (E3): R2(y1), R2(y1)
C (E2): R2(y1), R2(y1)

And I notice that my notation is not sufficient to describe all the state in Paxos. The Acceptors actually store two Epochs: the latest one promised, and the latest one accepted. Rephrasing the above, with the accepted and promised epochs shown separately: 

Primary P3 got Promises for E3 from A and B:
A (a:E1, p:E3): R1(x1), R2(x2), R3(x3)
B (a:E2, p:E3): R2(y1), R2(y1)
C (a:E2, p:E2): R2(y1), R2(y1)

Here "A (a:E1, p:E3)" means that safekeeper has accepted with E1, and promised E3. Now we can resolve the ambiguity. P3 replicates R2(y1) from B to A, because B had a higher accepted epoch than A. Final state is:

A (a:E3, p:E3): R1(x1), R2(y1)
B (a:E2, p:E3): R2(y1), R2(y1)
C (a:E2, p:E2): R2(y1), R2(y1)

hlinnaka: I'll work through this third scenario too:

> Other possible (really) situation:
> P1 performs R1_1(x1), R1_2(x2).
> A got all R1_1(x1), R1_2(x2).
> B got R1_1(x1)
> C also got R1_1(x1).
> P1 died.
> 
> P2 has votes for E2 from B and C. P2 performs R2_2(y1) and wrotes it to B and C. Then it died.
> 
> A has R1_1(x1), R1_2(x2)
> B has R1_1(x1), R2_2(y1)
> C has R1_1(x1).

The write to C isn't visible here. I think you mean that P2 writes R2_2(y1) to B only. I'll continue with that assumption.

> 
> P3 connects A and C for epoch E3. It replicates R1_2(x2) to C (either as R1_2(x2) or as R3_2(x2). And died.
> 
> A has R1_1(x1), R1_2(x2)
> B has R1_1(x1), R2_2(y1)
> C has R1_1(x1), R1_2(x2) (or R3_2(x2) if you like).
> 
> P4 connects to A and B for epoch E4. Which value it should replicate? x2 or y1?
> 
> Raft has an answer to this: R1_2(x2) is not considered as committed, because to consider it as committed P3 have to commit R3_3(z1) (ie write to both A and C).
> If it does, then R1_2(x2) is committed, and P4 have to replicate R1_2(x2) and R3_3(z1) to B.
> If it doesn't, then P4 have to replicate R2_2(y1) to A. Note, that it still will not be committed until P4 commits R4_3(thau1) as well.

So, rephrasing this scenario from the start with the notation that shows the accepted and promised epochs separately. Starting situation:

A(a:E1, p:E1): R1(x1), R2(x2)
B(a:E1, p:E1): R1(x1)
C(a:E1, p:E1): R1(x1)

P1 dies. P2 gets promises for E2 from B and C. P2 performs R2(y1) and writes it to B. Then it dies, too:

A(a:E1, p:E1): R1(x1), R2(x2)
B(a:E2, p:E2): R1(x1), R2(y1)
C(a:E1, p:E2): R1(x1)

P3 connects to A and C for epoch E3. It replicates R2(x2) to C And dies.

A(a:E1, p:E3): R1(x1), R2(x2)
B(a:E2, p:E2): R1(x1), R2(y1)
C(a:E3, p:E3): R1(x1), R2(x2)

P4 connects to A and B for epoch E4

A(a:E1, p:E4): R1(x1), R2(x2)
B(a:E2, p:E4): R1(x1), R2(y1)
C(a:E3, p:E3): R1(x1), R2(x2)

> Which value it should replicate? x2 or y1?
>
> Raft has an answer to this: R1_2(x2) is not considered as committed, because to consider it as committed P3 have to commit R3_3(z1) (ie write to both A and C).
> If it does, then R1_2(x2) is committed, and P4 have to replicate R1_2(x2) and R3_3(z1) to B.
> If it doesn't, then P4 have to replicate R2_2(y1) to A. Note, that it still will not be committed until P4 commits R4_3(thau1) as well.

Gotcha. Yeah, with Paxos too, R2(x2) is not considered as committed until the accepted value has been updated in a majority of acceptors. So in this case, no value has been committed for R2 yet, and replicating either R2(x2) or R2(y1) would be valid. B has a higher accepted Epoch value, though, so P4 will choose to replicate R2(y1).

A(a:E4, p:E4): R1(x1), R2(y1)
B(a:E2, p:E4): R1(x1), R2(y1)
C(a:E3, p:E3): R1(x1), R2(x2)

R2(y1) still cannot be considered as committed, until B has accepted the new epoch too. So as final step, P4 needs to send an Accept message to B:

A(a:E4, p:E4): R1(x1), R2(y1)
B(a:E4, p:E4): R1(x1), R2(y1)
C(a:E3, p:E3): R1(x1), R2(x2)

Now R2(y1) is committed.

hlinnaka: > Raft is not separate algorithm, as one might think. Raft is well and thoroughly described MultiPaxos as it should be (ok, with some tuning: placing Proposer, Accepter and Learner at same processes). One have to understand quirks of Raft to clearly understand MultiPaxos.

Sure.

funny-falcon: > Here "A (a:E1, p:E3)" means that safekeeper has accepted with E1, and promised E3. Now we can resolve the ambiguity. P3 replicates R2(y1) from B to A, because B had a higher accepted epoch than A.

Lets use your terminology.
By the way, then "safekeeper had accepted epoch" happens? Does it happens when WAL record is written? Or it is separate message not bound to WAL?

Lets assume it is bound to WAL record:

P1 performs R1(x1), R1(x2), R1(x3). A got all R1(x1), R1(x2) and R1(x3). B got R1(x1), C also got R1(x1). P1 dies. Situation:

A (E1:E1): R1(x1), R2(x2), R3(x3)
B (E1:E1): R1(x1)
C (E1:E1): R1(x1)

P2 has votes for E2 from B and C. P2 performs R2(y1) and writes it to B. Then it dies. Now we have:
A (E1:E1): R1(x1), R2(x2), R3(x3)
B (E2:E2): R1(x1), R2(y1)
C (E1:E2): R1(x1)

Primary P3 gets Promises for E3 from A and C. It safely replicates R2(x2) to C. Then it dies.
A (E1:E3): R1(x1), R2(x2), R3(x3)
B (E2:E2): R1(x1), R2(y1)
C (E3:E3): R1(x1), R2(x2)

Primary P4 gets promises for E4 from A and B. B had accepted E2, A has accepted E1. Which value P4 should it replicate, x2 or y1?

*Text below is not importand for discussion. Skip it.*

Now assume "safekeeper had accepted epoch" is a separate process that happens before new (from current safekeeper point of view) WAL records are written.

Well, if proposer (ie primary) waits for acceptance of epoch from majority, then it looks like there no loss of committed records.

For example:
P1 performs R1(x1), R1(x2). A got all R1(x1), R1(x2). B got R1(x1), C also got R1(x1). P1 dies. Situation:

A (E1:E1): R1(x1), R2(x2)
B (E1:E1): R1(x1)
C (E1:E1): R1(x1)

P2 got election (ie promise + accept) from B and C. Then it replicates R2(y1) to B.
A (E1:E1): R1(x1), R2(x2)
B (E2:E2): R1(x1), R2(y1)
C (E2:E2): R1(x1)

Now P3 got election from A and C. Since C has accepted epoch E2, P3 has to truncate log of A to record R1(x1) first:
A (E3:E3): R1(x1)
B (E2:E2): R1(x1), R2(y1)
C (E3:E3): R1(x1)
Then it can write record R2(z1) to A:
A (E3:E3): R1(x1), R2(z1)
B (E2:E2): R1(x1), R2(y1)
C (E3:E3): R1(x1)

Now P4 elected by B and C. Since C has accepted epoch E3, P4 has to truncate B log to R1(x1):
A (E3:E3): R1(x1), R2(z1)
B (E4:E4): R1(x1)
C (E4:E4): R1(x1)

Hey, oops. What if P3 replicated R2(z1) to both A and C ??? Is it committed?
A (E3:E3): R1(x1), R2(z1)
B (E2:E2): R1(x1), R2(y1)
C (E3:E3): R1(x1), R2(z1)

Now P4 elected by B and C and truncates B log (cause it will need to add R2(z1). And then fails.
A (E3:E3): R1(x1), R2(z1)
B (E4:E4): R1(x1)
C (E4:E4): R1(x1), R2(z1)

Next P5 is elected by A and B. Now it have to truncate A log. Therefore R2(z1) became "uncommitted".

Assume we do not truncate log, but overwrite. Then it needs additional record to make mistake. Lets assume after P3 dies we get:
A (E3:E3): R1(x1), R2(z1), R3(z2)
B (E2:E2): R1(x1), R2(y1), R3(y2)
C (E3:E3): R1(x1), R2(z1), R3(z2)

Then P4 elected by B and C. It starts to overwrite B log, but died after overwrite of R2.
A (E3:E3): R1(x1), R2(z1), R3(z2)
B (E4:E4): R1(x1), R2(z1), R3(y2)
C (E4:E4): R1(x1), R2(z1), R3(z2)

Now P5 is elected by A and B. B has higher accepted epoch, therefore P5 replicates its log to A:
A (E5:E5): R1(x1), R2(z1), R3(y2)
B (E5:E5): R1(x1), R2(z1), R3(y2)
C (E4:E4): R1(x1), R2(z1), R3(z2)

Oops, we lost committed z2 value.

funny-falcon: > Gotcha. Yeah, with Paxos too, R2(x2) is not considered as committed until the accepted value has been updated in a majority of acceptors. So in this case, no value has been committed for R2 yet, and replicating either R2(x2) or R2(y1) would be valid. B has a higher accepted Epoch value, though, so P4 will choose to replicate R2(y1).

Oh, i've missed this comment before I wrote previous message. Yes, with "Paxos doesn't consider ... as committed until ..." it becames correct consensus. Then in my first example in previous message P4 is safe to truncate A log to R1, because R2(x2) is not considered as committed.

Pain point will be to find common log record. Epoch in WAL record really helps to do it. Otherwise some other equality check should be invented.

PR #4, opened 2021-03-29
zenith-local design draft

lubennikovaav:
PR for discussion
Based on the discussion in https://github.com/libzenith/rfcs/pull/2

### DIFF
--- a/zenith_local.md
+++ b/zenith_local.md
@@ -1,0 +1,103 @@
+# Zenith local
+
+Here I list some objectives to keep in mind when discussing zenith-local design and a proposal that brings all components together.  Your comments on both parts are very welcome.
+
+#### Why do we need it?
+- For distribution - this easy to use binary will help us to build adoption among developers.
+- For internal use - to test all components together.
+
+In my understanding, we consider it to be just a mock-up version of zenith-cloud.
+> Question: How much should we care about durability and security issues for a local setup?
+
+
+#### Why is it better than a simple local postgres?
+
+- Easy one-line setup. As simple as `cargo install zenith && zenith start`
+
+- Quick and cheap creation of compute nodes over the same storage.
+> Question: How can we describe a use-case for this feature?
+
+- Zenith-local can work with S3 directly. 
+
+- Push and pull images (snapshots) to remote S3 to exchange data with other users.
+
+- Quick and cheap snapshot checkouts to switch back and forth in the database history.
+> Question: Do we want it in the very first release? This feature seems quite complicated.
+
+#### Distribution:
+
+Ideally, just one binary that incorporates all elements we need.
+> Question: Let's discuss pros and cons of having a separate package with modified PostgreSQL.
+
+#### Components:
+
+- **zenith-CLI** - interface for end-users.  Turns commands to REST requests and handles responces to show them in a user-friendly way.  
+CLI proposal is here https://github.com/libzenith/rfcs/blob/003-laptop-cli.md/003-laptop-cli.md
+WIP code is here: https://github.com/libzenith/postgres/tree/main/pageserver/src/bin/cli
+
+- **zenith-console** - WEB UI with same functionality as CLI.
+>Note: not for the first release.
+
+- **zenith-local** - entrypoint. Service that starts all other components and handles REST API requests. See REST API proposal below.
+    > Idea: spawn all other components as child processes, so that we could shutdown everything by stopping zenith-local.
+
+- **zenith-pageserver** - consists of a storage and WAL-replaying service (modified PG in current implementation).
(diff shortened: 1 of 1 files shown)


zenith_local.md
```diff
@@ -0,0 +1,103 @@
...
+- For distribution - this easy to use binary will help us to build adoption among developers.
+- For internal use - to test all components together.
+
+In my understanding, we consider it to be just a mock-up version of zenith-cloud.
+> Question: How much should we care about durability and security issues for a local setup?
+
+
+#### Why is it better than a simple local postgres?
+
+- Easy one-line setup. As simple as `cargo install zenith && zenith start`
```
kelvich, line 15: Actually that probably wouldn't -- I researched topic. Cargo deliberately prohibits a lot actions needed for package manager like pre/post scripts, included files, etc. If software isn't completely in rust (and we have to carry at least postgres) it is hard to use cargo.

zenith_local.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Quick and cheap creation of compute nodes over the same storage.
+> Question: How can we describe a use-case for this feature?
```
kelvich, line 18: People who already stuck with k8s would love that.

zenith_local.md
kelvich: So we may use brew (mac), choco (win), snap (linux)

zenith_local.md
```diff
@@ -0,0 +1,103 @@
...
+#### REST API:
+
+Service endpoint: `http://localhost:3000`
+
+Resources:
+- /storages - Where data lives: zenith-pageserver or zenith-s3
+- /pgs - Postgres - zenith-computenode
+- /snapshots - snapshots **TODO**
+
+>Question: Do we want to extend this API to manage zenith components? I.e. start page-server, manage safekeepers and so on? Or they will be hardcoded to just start once and for all?
```
kelvich, line 73: I would try to consider them static for both local and cloud installation (probably except failure recovery scenario in cloud)

kelvich:
Thanks for writing!

Let's try also invent here some small-step milestones to avoid working on everything simultaneously. First milestone could have following functionality:

zenith start/stop -- start/stop zenithd
zenith postgres init/start/stop -- manage databases

zenithd is a service that starts threads for walkeeper, page server, entrypoint. Page server still stores data in-memory, walkeeper never trims the wal.
zenith communicates with zenithd using http. Important part here is protocol between them.
zenith stop destroys data directory of corresponding postgres, so that postgres start would need to deal somehow with that.
zenithd has toml config that specifies what addresses/roles it starts and path to postgres binaries. That would be enough to try to pack it with brew/snap and test distribution process itself.

On the second milestone we add S3 and persistency to this scheme.
On the third we add snapshot with various transfer/import/export options.

zenith_local.md
```diff
@@ -0,0 +1,103 @@
...
+storage attach -n name --type [native\s3]  --path=[datadir\URL] | PUT  -d { "name": "name", "type": "native", "path": "/tmp" } /storages
+storage detach -n name | DELETE /storages/:storage_name 
+storage list | GET /storages
+storage show -n name | GET /storages/:storage_name 
+
+
+- /pgs - zenith-computenode
+
+CLI  | REST API
+------------- | -------------
```
kelvich, line 90: Most of that API calls may fail with variety of reasons. So I would better not model pg as a REST service, but instead have a REST-like api for actions with postgres. Like following:

PUT {"action":"pg.start", ...}  -> {"action_id": 12345}
GET  {"action_id": 12345} -> {"status": "in-progress"}
GET  {"action_id": 12345} -> {"status": "in-progress"}
GET  {"action_id": 12345} -> {"status": "miserably-failed", "log": "..."}

So that in case of error we were able to return some details, not 500 code. I assume CLI could ask status each X ms, and wait until status is final. In case of web console we may move all that communication into websocket and send status from zenithd when it is ready.

zenith_local.md
hlinnaka: We can use a Makefile and `make install`. It can call `cargo` for all the Rusty parts.

zenith_local.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Why do we need it?
+- For distribution - this easy to use binary will help us to build adoption among developers.
+- For internal use - to test all components together.
+
+In my understanding, we consider it to be just a mock-up version of zenith-cloud.
+> Question: How much should we care about durability and security issues for a local setup?
+
+
+#### Why is it better than a simple local postgres?
```
knizhnik, line 13: I think that there will  can provide at least two more advantages comparing with vanilla Postgres:
page-level compression and encryption. We can quite easily add them, unlike standard Postgres.
Encryption requires more attention and efforts than compression because we need to concern encryption of WAL and local file cache (if any).

PR #5, opened 2021-04-09
CLI v2 for discussion

hlinnaka:
@nikitashamgunov wrote this, I copied it from Notion for easier commenting

### DIFF
--- a/repository-structure.md
+++ b/repository-structure.md
@@ -1,0 +1,140 @@
+# Repository format
+
+A Zenith repository is similar to a traditional PostgreSQL backup
+archive, like a WAL-G bucket or pgbarman backup catalogue. It holds
+multiple versions of a PostgreSQL database cluster.
+
+The distinguishing feature is that you can launch a Zenith Postgres
+server directly against a branch in the repository, without having to
+"restore" it first. Also, Zenith manages the storage automatically,
+there is no separation between full and incremental backups nor WAL
+archive. Zenith relies heavily on the WAL, and uses concepts similar
+to incremental backups and WAL archiving internally, but it is hidden
+from the user.
+
+## Directory structure, version 1
+
+This first version is pretty straightforward but not very
+efficient. Just something to get us started.
+
+The repository directory looks like this:
+
+    .zenith/timelines/4543be3daeab2ed4e58a285cbb8dd1fce6970f8c/wal/
+    .zenith/timelines/4543be3daeab2ed4e58a285cbb8dd1fce6970f8c/snapshots/<lsn>/
+    .zenith/timelines/4543be3daeab2ed4e58a285cbb8dd1fce6970f8c/history
+    
+    .zenith/refs/branches/mybranch
+    .zenith/refs/tags/foo
+    .zenith/refs/tags/bar
+    
+    .zenith/datadirs/<timeline uuid>
+
+### Timelines
+
+A timeline is similar to PostgeSQL's timeline, but is identified by a
+UUID instead of a 32-bit timeline Id.  For user convenience, it can be
+given a name that refers to the UUID (called a branch).
+
+All WAL is generated on a timeline. You can launch a read-only node
+against a tag or arbitrary LSN on a timeline, but in order to write,
+you need to create a timeline.
+
+Each timeline is stored in a directory under .zenith/timelines. It
+consists of a WAL archive, containing all the WAL in the standard
+PostgreSQL format, under the wal/ subdirectory.
+
+The 'snapshots/' subdirectory, contains "base backups" of the data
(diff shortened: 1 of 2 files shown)


hlinnaka:
Overall, I like this.

Some comments:
- "commit" is a loaded word in database world. Need to invent a different name for that
- a PostgreSQL is never truly read-only, unless it's started as a read-only replica. Even read-only queries often modify pages, to set hint bits and do other cleanup tasks that are not directly visible to users. But they generate WAL. And autovacuum runs in the background at unpredictable times and generates WAL. So as soon as you start the instance, you will have made "changes" that show up in `zenith status`.

Wrt. implementing this:
- this changes our focus from the WAL safekeeper and Page Server. If you run this locally, you don't really want to manage those extra processes, they should be integrated so you only need to launch Postgres. So I'm thinking that this should all be handled by a new extension that hooks into the Storage Manager API, and there is no WAL safekeeper nor Page Server involved on your laptop.
- Do will still want to have the WAL safekeeper and Page Server for when you launch an instance in the cloud? The alternative is to point all the compute nodes directly at the S3 bucket. I think we do want them though. WAL safekeeper gives you the extra durability, if the compute node is hit by a meteor. And the page servers allow sharing resources between compute nodes, if you have read-only nodes.
- The file format that we will use for the repository is really crucial here. It should be non-overwriting, like in git. But also allow efficient garbage collection of all the page versions that are no longer needed.

hlinnaka:
On our team meeting call yesterday, IIRC Nikita asked why the Page Server need to have a custom page cache implementation, why cannot we just use the PostgreSQL's buffer cache. I said because the Page Server needs to hold multiple versions of each page, and be prepared to return any page if needed, for the read-only replicas. That concern goes away in the local extension. If you have a Postgres extension that works directly with the repository file format, that extension only needs to reconstruct the "current" page version for the Postgres instance, not arbitrary versions. So the extension doesn't need a page cache of its own.

hlinnaka:
> "commit" is a loaded word in database world. Need to invent a different name for that

I think these are equivalent to what Stas has been calling "snapshots". "Snapshot" seems good to me.

hlinnaka:
Hosting public repositories that anyone can check out and run locally on a laptop has some security implications. It's easy to create a cluster that does nasty things when you launch a Postgres instance running it.

The attack here is:
1. Mallory creates a database of cute cat images.
2. Mallory also creates a trigger or something that launches an ssh server as soon as you SELECT a table.
3. Mallory upload the database to the cloud and posts a link to it to reddit.
4. Alice runs `zenith clone zenith://zenith.tech/catphotos` and `zenith start`s a local instance.
5. Alice runs `SELECT * FROM cat_photos`. That launches the ssh server and Mallory steals her data.

The bottom line is that you have to trust the author of the repository.

hlinnaka:
I added a new file on this RFC branch for a proposed directory layout for the Zenith repository, to support these CLI commands. Please take a look and comment

https://github.com/zenithdb/rfcs/blob/laptop-cli-v2/repository-structure.md

repository-structure.md
```diff
@@ -0,0 +1,140 @@
...
+contains the UUID of the timeline (and LSN, for tags).
+
+### Datadirs
+
+.zenith/datadirs contains PostgreSQL data directories. You can launch
+a Postgres instance on one of them with:
+
+```
+  postgres -D .zenith/datadirs/4543be3daeab2ed4e58a285cbb8dd1fce6970f8c
+```
```
kelvich, line 74: So in this concept zenith-cli launches and manages postgres itself? Does this imply having some zenithd constantly running? Or cli is passive and it prepares data directory where user can start postgres?

kelvich:
Also can we somehow marry that concept with client (bgworker or standalone daemon) for streaming local postgres to remote walkeeper?

repository-structure.md
hlinnaka: I'm thinking that the "zenith start" command prepares the data directory, and calls `pg_ctl -D .zenith/datadirs/4543be3daeab2ed4e58a285cbb8dd1fce6970f8c start`. No need to have a zenithd running. We can have a "zenith list running intances" command and "zenith stop/restart", but they can all work by looking at the pid files in the data directories and calling "pg_ctl".

kelvich: On a second read it seems that cli is meant to be passive

hlinnaka:
> Also can we somehow marry that concept with client (bgworker or standalone daemon) for streaming local postgres to remote walkeeper?

Hmm. In that case, the repository would be remote, but the data directory would be local. In that case, the extension would connect to the remote walkeeper and page server. The remote walkeeper/page server is then responsible for reading/updating the repository. The Page Server needs to do the same things that the local extension does in the local case. We should try to use the same code for both somehow..

PR #6, opened 2021-05-05
RFC: Serverless on your laptop.

hlinnaka:
What if we automate the "zenith pg create" and "zenith pg start" steps into
the Page Server, so that it becomes Serverless? Serverless here mean that
you only need to start the Page Server, not the individual Postgres
instances.

### DIFF
--- a/serverless-on-laptop.md
+++ b/serverless-on-laptop.md
@@ -1,0 +1,93 @@
+How it works now
+----------------
+
+1. Create repository, start page server on it
+
+```
+$ zenith init
+...
+created main branch
+new zenith repository was created in .zenith
+
+$ zenith pageserver start
+Starting pageserver at '127.0.0.1:64000' in .zenith
+Page server started
+```
+
+2. Create a branch, and start a Postgres instance on it
+
+```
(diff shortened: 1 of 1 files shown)


serverless-on-laptop.md
```diff
@@ -0,0 +1,93 @@
...
+
+1. Create repository, start page server on it (same as before)
+
+```
+$ zenith init
+...
+created main branch
+new zenith repository was created in .zenith
+
+$ zenith pageserver start
```
ericseppanen, line 60: Does a user care what a "pageserver" is? Maybe this could just be "zenith start"?

ericseppanen:
Another thought: should it work like docker?  I can type `docker run -it ubuntu:18.04` and I have a running container that "branched off" from the upstream Ubuntu image.

I imagine I could type `zenith start --empty` which starts an instance with no content. Or I could type `zenith start mydb_snapshot_yesterday` or `zenith start mycompany.com/db43/2021_04_05` to start an instance that is a new branch from a known point. Known to some local naming scheme or repository service, at least.

PR #7, opened 2021-05-11
push-pull

kelvich:
[rendered](https://github.com/zenithdb/rfcs/blob/push-pull/push-pull.md)

### DIFF
--- a/push-pull.md
+++ b/push-pull.md
@@ -29,22 +29,22 @@ Exact mechanics would be slightly different in the following situations:
 
 1) Destination branch does not exist.
 
-That is the simplest scenario. We can just create an empty branch (or timeline in internal terminology) and transfer all the pages/records that we have in our timeline. Right now each timeline is quite independent of other timelines so I suggest skipping any checks that there is a common ancestor and just fill it with data. Later when CoW timelines will land to the pageserver we may add that check and decide whether this timeline belongs to this pageserver repository or not [*].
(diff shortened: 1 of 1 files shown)


push-pull.md
```diff
@@ -0,0 +1,66 @@
...
+    * Sending pageserver may start iterator on that timeline and send all the records as copy messages.
+
+2) Destination branch exists and latest_valid_lsn is less than ours.
+
+    In this case, we need to send missing records. To do that we need to find all pages that were changed since that remote LSN. Right now we don't have any tracking mechanism for that, so let's just iterate over all records and send ones that are newer than remote LSN. Later we probably should add a sparse bitmap that would track changed pages to avoid full scan.
+
+3) Destination branch exists and latest_valid_lsn is bigger than ours.
+
+    In this case, we can't push to that branch. We can only pull.
+
```
hlinnaka, line 48: A more problematic case is:

4) Destination branch exists, but with different content
For example, the branch identical in the past, but new WAL records were added it in the local repository, and in the remote repository. But they're different.

kelvich: True, missed that one. We could use blockchain-like approach for detecting this situation: maintain last_wal_hash = hash( last_wal_record . last_wal_hash ). By comparing them we can deduce that branches are same. That wouldn't work in case when push/pull was interrupted since we modify data without general WAL. And interrupted pull may be handled differently, e.g. we can be explicit about pull start and save this info on destination; pull retry may check that information.

PR #8, opened 2021-06-17
RFC for "snapshot-first storage" architecture

ericseppanen:
RFC for a storage architecture that uses s3 as the primary home for snapshots, and snapshots as the primary way of permanently storing data.

[Rendered](https://github.com/zenithdb/rfcs/blob/7810fcadbcd140241cd595304df1eeb7a7dd6718/snapshot-first-storage.md)

### DIFF
--- a/snapshot-first-storage.md
+++ b/snapshot-first-storage.md
@@ -1,0 +1,148 @@
+# Snapshot-first storage architecture
+
+Goals:
+- Long-term storage of database pages.
+- Easy snapshots; simple snapshot and branch management.
+- Allow cloud-based snapshot/branch management.
+- Allow cloud-centric branching; decouple branch state from running pageserver.
+- Allow customer ownership of data via s3 permissions.
+- Provide same or better performance for typical workloads, vs plain postgres.
+
+Non-goals:
+- Service database reads from s3 (reads should be serviced from the pageserver cache).
+- Keep every version of every page / Implement point-in-time recovery (possibly a future paid feature, based on WAL replay from an existing snapshot).
+
+## Principle of operation
+
+The database “lives in s3”. This means that all of the long term page storage is in s3, and the “live database”-- the version that lives in the pageserver-- is a set of “dirty pages” that haven’t yet been written back to s3.
+
+In practice, this is mostly similar to storing frequent snapshots to s3 of a database that lives primarily elsewhere.
+
+The main difference is that s3 is authoritative about which branches exist; pageservers consume branches, snapshots, and related metadata by reading them from s3. This allows cloud-based management of branches and snapshots, regardless of whether a pageserver is running or not.
+
+It’s expected that a pageserver should keep a copy of all pages, to shield users from s3 latency. A cheap/slow pageserver that falls back to s3 for some reads would be possible, but doesn’t seem very useful right now.
+
+Because s3 keeps all history, and the safekeeper(s) preserve any WAL records needed to reconstruct the most recent changes, the pageserver can store dirty pages in RAM or using non-durable local storage; this should allow very good write performance, since there is no need for fsync or journaling.
+
+Objects in s3 are immutable snapshots, never to be modified once written (only deleted).
+
+Objects in s3 are files, each containing a set of pages for some branch/relation/segment as of a specific time (LSN). A snapshot could be complete (meaning it has a copy of every page), or it could be incremental (containing only the pages that were modified since the previous snapshot). It’s expected that most snapshots are incremental to keep storage costs low.
+
+It’s expected that the pageserver would upload new snapshot objects frequently, e.g. somewhere between 30 seconds and 15 minutes, depending on cost/performance balance.
+
+No-longer needed snapshots can be “squashed”-- meaning snapshot N and snapshot N+1 can be read by some cloud agent software, which writes out a new object containing the combined set of pages (keeping only the newest version of each page) and then deletes the original snapshots.
+
+A pageserver only needs to store the set of pages needed to satisfy operations in flight: if a snapshot is still being written, the pageserver needs to hold historical pages so that snapshot captures a consistent moment in time (similar to what is needed to satisfy a slow replica).
+
+WAL records can be discarded once a snapshot has been stored to s3. (Unless we want to keep them longer as part of a point-in-time recovery feature.)
+
+## Pageserver operation
+
+To start a pageserver from a stored snapshot, the pageserver downloads a set of snapshots sufficient to start handling requests. We assume this includes the latest copy of every page, though it might be possible to start handling requests early, and retrieve pages for the first time only when needed.
+
+To halt a pageserver, one final snapshot should be written containing all pending WAL updates; then the pageserver and safekeepers can shut down.
+
+It’s assumed there is some cloud management service that ensures only one pageserver is active and servicing writes to a given branch.
+
+The pageserver needs to be able to track whether a given page has been modified since the last snapshot, and should be able to produce the set of dirty pages efficiently to create a new snapshot.
+
+The pageserver need only store pages that are “reachable” from a particular LSN. For example, a page may be written four times, at LSN 100, 200, 300, and 400. If no snapshot is being created when LSN 200 is written, the page at LSN 100 can be discarded. If a snapshot is triggered when the pageserver is at LSN 299, the pageserver must preserve the page from LSN 200 until that snapshot is complete. As before, the page at LSN 300 can be discarded when the LSN 400 pages is written (regardless of whether the LSN 200 snapshot has completed.)
+
+If the pageserver is servicing multiple branches, those branches may contain common history. While it would be possible to serve branches with zero knowledge of their common history, a pageserver could save a lot of space using an awareness of branch history to share the common set of pages. Computing the “liveness” of a historical page may be tricky in the face of multiple branches.
+
+The pageserver may store dirty pages to memory or to local block storage; any local block storage format is only temporary “overflow” storage, and is not expected to be readable by future software versions.
+
+The pageserver may store clean pages (those that are captured in a snapshot) any way it likes: in memory, in a local filesystem (possibly keeping a local copy of the snapshot file), or using some custom storage format. Reading pages from s3 would be functional, but is expected to be prohibitively slow.
+
+The mechanism for recovery after a pageserver failure is WAL redo. If we find that too slow in some situations (e.g. write-heavy workload causes long startup), we can write more frequent snapshots to keep the number of outstanding WAL records low. If that’s still not good enough, we could look at other options (e.g. redundant pageserver or an EBS page journal).
+
+A read-only pageserver is possible; such a pageserver could be a read-only cache of a specific snapshot, or could auto-update to the latest snapshot on some branch. Either way, no safekeeper is required. Multiple read-only pageservers could exist for a single branch or snapshot.
+
+## Cloud snapshot manager operation
+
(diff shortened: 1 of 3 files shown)


hlinnaka:
Looks good to me!

> Snapshot files (and any other object stored in s3) must be readable by future software versions.

I agree. It occurred to me, though, that there's another way we could handle upgrades: the snapshot files are immutable, so we could have an offline tool that reads the old format and writes the new format. With that, you could create a copy of an existing S3 bucket in the new format, and launch new page servers against that copy, while the old ones are still running. As long as the old and the new versions can digest the WAL stream coming from the walkeeper, you can have two page server versions running side-by-side for as long as you want.

I'm not sure if that's easier than being able to read the old format. But it's another option we have.

knizhnik:
Some comments from my side:
1. Should we separate branches and snapshots/backups? At least at API level?  Branches are used to be created explictly by user, while snapshots are most frequently taken automatically (for example each day). From implementation point of view switching timeline for each snapshot (i.e. once per  day) seems to be not so good behavior.
2. How we should handle incremental backups? This question is somehow elated to 1). If we have to snapshots S1 and S2, then S1 is not an ancestor of S2, Both are descendants of main branch.  But we what to store S2 as delta to S1. I am not sure how sharing version history between S1, S2 and main branch will help here...
3. What is the smallest unit of changes in snapshot? 8kb page?  But what if just one record is changed on this page? So should we consider storing deltas (wal record) for some pages instead of full images in snapshots. It certainly significantly complicates recovery but can save a lot of space...  I am not sure here, pg_probackup is working only at page level and may be we should also start from this. 
4. Compression of pages in snashots/backups is "must have" feature, But should we do it ourselves or rely on S3? In the last case we will have to pay for extra traffic between pageserver and S3.

ericseppanen:
I don't think there's any difference between snapshots, backups, and branches in the format of the stored file. I expect that the only way branches are treated differently is that the cloud API should make it easy to start a pageserver on that branch in read/write mode, and should not allow starting a pageserver in read/write mode on a random snapshot (a new branch should be forked from that snapshot instead).

Snapshots are defined as a collection of 8KB pages. If one byte has changed, then the full page will be stored in the snapshot. Snapshots are basically a page storage container, so the format is simple and reading them is also simple. I hope we don't need to store WAL records in them, since that would require starting a copy of postgres to be able to squash two snapshots together. The history of any branch is linear: each snapshot has exactly one parent snapshot, and an incremental snapshot contains changes relative to its parent.

I don't think that compression needs to be part of the file format for v1. I don't object to compression, but it's mostly invisible to customers so I don't think it's on the critical path for the first release. Most snapshot files will only be kept for a short time (hours or days) before being squashed into another snapshot, so compression is probably wasted effort until we know that the snapshots will be stored for a long time. This could be a cloud snapshot manager policy-- e.g. when a snapshot is more than 7 days old, compress the entire file.

snapshot-first-storage.md
```diff
@@ -0,0 +1,148 @@
...
+
+To halt a pageserver, one final snapshot should be written containing all pending WAL updates; then the pageserver and safekeepers can shut down.
+
+It’s assumed there is some cloud management service that ensures only one pageserver is active and servicing writes to a given branch.
+
+The pageserver needs to be able to track whether a given page has been modified since the last snapshot, and should be able to produce the set of dirty pages efficiently to create a new snapshot.
+
+The pageserver need only store pages that are “reachable” from a particular LSN. For example, a page may be written four times, at LSN 100, 200, 300, and 400. If no snapshot is being created when LSN 200 is written, the page at LSN 100 can be discarded. If a snapshot is triggered when the pageserver is at LSN 299, the pageserver must preserve the page from LSN 200 until that snapshot is complete. As before, the page at LSN 300 can be discarded when the LSN 400 pages is written (regardless of whether the LSN 200 snapshot has completed.)
```
ololobus, line 49: Regarding:

> If no snapshot is being created when LSN 200 is written, the page at LSN 100 can be discarded.
> ...
> As before, the page at LSN 300 can be discarded when the LSN 400 pages is written (regardless of whether the LSN 200 snapshot has completed.)

Is it completely unrealistic for compute node (Postgres) to ask Page@LSN100 even if it just wrote WAL records for Page@LSN200?

I thought that due to MVCC it is legal that some long running transaction with an older snapshot can read older version of some page. And this is not that rare when a lot of concurrent transactions occur. So in our case, if Pageserver just drops all images of PageN with LSN < 200 when materializing/writing Page@LSN200, then on the next GetPage@LSN100 we should go to S3 and materialize this older page since the prior snapshot. If so, is it fine from the performance (latency) perspective?

ololobus: On the second thought, I mixed a concept of MVCC with different versions of **record** and different **pages**. So if we write a new record version we may create a new page for doing that, but these are completely different pages from the Pageserver perspective and everything is fine. Sorry for inconvenience :)

knizhnik: There is actually no MVCC at page level. Long living transaction will still read page with most recent LSN (actually with last written LSN, but it doesn't matter).   So master will never try to read old page version even in case of presence of long living transactions. It may be true only for lagged replicas.

hlinnaka: > On the second thought, I mixed a concept of MVCC with different versions of **record** and different **pages**.

Yeah, MVCC is not an issue here.

You have a point though. Currently, the compute node *will* in fact always use a slightly lagging-behind LSN in the requests. It keeps track of the LSN of the last page that was evicted from the buffer cache, and uses that in the requests, instead of the LSN of the last-inserted WAL record as you might expect. That's an optimization. We know that the page being requested cannot have changed after the last-evicted LSN. If it was, it would either still be in the buffer cache and we would not be requesting it from the page server, or we would have updated the last-evicted LSN. That optimization allows the page server to respond immediately to the GetPage@LSN request even if it hasn't fully caught up to the latest generated LSN, reducing latency.

But that could be a problem if the page server had already garbage collected some page versions after the last-evicted LSN. Perhaps the GetPage@LSN request should include two LSNs, min and max. The semantics would be that the compute node knows that the page hasn't changed between the min and max LSN, and the page server is therefore allowed to return any page version in that range (there should be only one, we could make it an error if there are multiple). The compute node would use the last-evicted LSN as the min, and last-generated LSN as the max.

Alternatively, the GetPage@LSN request could carry a flag to indicate that a later LSN is OK as well. So the page server could respond with a page at the given LSN or later (and there should again be only one such version in practice).

ololobus: That is interesting, thanks for the explanation!

lubennikovaav:
I've added one more RFC that fits import/export commands to 'snapshot-first' design and focuses on UI design that is not covered in the first RFC.
[Rendered](https://github.com/zenithdb/rfcs/blob/storage_architecture/snapshot-storage-cli.md)

hlinnaka:
I wrote down some thoughts comparing what we currently have in the 'layered-repo' branch and this RFC, and some ideas on how to proceed.

Rendered here: https://github.com/zenithdb/rfcs/blob/storage_architecture/snapshot-storage-pitr.md

@ericseppanen

ericseppanen:
I'm OK with the storage format allowing storage of WAL records; I don't know that it should be required or even the default.

If WAL records do captured in a snapshot file, I think they should be "the history leading up to" this snapshot, i.e. the snapshot covering LSN 100-200 should contain pages as of LSN 200, and any WAL records present should cover 100-200. A few reasons why:
- A cloud control plane or snapshot manager should be able to manage snapshots without needing to spin up a the right version of postgres to run WAL redo.
- It would be cumbersome to spawn a new branch from an old snapshot, if WAL redo is needed just to catch up to the snapshot's latest LSN.
- The pageserver would need to keep all of the "previous" snapshot's pages in memory until the next snapshot, which seems backwards: if the new snapshot contains the latest pages, then it's possible to drop those from memory as soon as the snapshot is persistent, and only accumulate changes since then.

I'm somewhat afraid of any design decisions that would make the snapshot files heavier. If the snapshots are too heavy, then the snapshot frequency will need to go down, restarts will be slow (because of longer WAL redo), user requests (e.g. branching) will be slow, etc. I would wish for a cloud UI that is responsive and gives the impression that all the data is instantly synchronized.

kelvich:
Some more thought on storage format: https://github.com/zenithdb/rfcs/blob/storage_details/storage_details.md

petergeoghegan:
> If WAL records do captured in a snapshot file, I think they should be "the history leading up to" this snapshot, i.e. the snapshot covering LSN 100-200 should contain pages as of LSN 200, and any WAL records present should cover 100-200.

In general WAL records are not useful without a base page image to apply them against. Whatever the merits of storing WAL records alongside pages in snapshots may be, I can't see what purpose this approach would serve. Perhaps I missed something.

> * A cloud control plane or snapshot manager should be able to manage snapshots without needing to spin up a the right version of postgres to run WAL redo.

The data isn't meaningful without being interpreted by Postgres anyway. So why does this matter? Can't the snapshot manager be taught that a base image plus some additional WAL records map to a certain *logical* LSN?

> * It would be cumbersome to spawn a new branch from an old snapshot, if WAL redo is needed just to catch up to the snapshot's latest LSN.

I agree that it's important that the volume of WAL be carefully bounded in order to minimize *perceptible* recovery time, whatever that may end up meaning. A bit like a standard Postgres backup. Though unlike standard Postgres, it might be okay to have lots of WAL records, once we can be sure that they were not concentrated in any one relation or page. Possibly through some kind of backpressure mechanism. This likely cannot be done purely inside the snapshot manager.

> I'm somewhat afraid of any design decisions that would make the snapshot files heavier.

Keeping the snapshot format simple seems important. But it's possible that storing some WAL records in some fashion will make life easier, not harder. Postgres basebackups more or less work by leveraging WAL. As a result, Postgres doesn't need to be quiescent (in any sense) during the backup. Plus the third party backup tool doesn't have to care about file-level consistency provided it follows some simple high level rules. This gives the backup tool significant freedom. It has the freedom to not have to worry about filesystem-level consistency in detail, provided the tool follows certain high level rules.

Might be useful to review the docs for the low-level continuous archiving functions in Postgres, to get a rough sense of how this works:

https://www.postgresql.org/docs/devel/continuous-archiving.html#BACKUP-LOWLEVEL-BASE-BACKUP

Allowing the snapshot format to have some limited understanding of WAL records/logical LSNs/whatever might actually *enable* cheap branching that users will find compelling. Users naturally think of their database in high level terms, and I suspect that tolerating "non-materialized pages" in snapshots will help with that.

Users will find it compelling to branch using a built-in function, that they may run from an interactive psql session. I'm thinking of something a little like pg_stop_backup() (from the docs I linked to). pg_stop_backup() returns an LSN that is the exact point that the final backup will restore to (at least in simple cases without pitr). It could easily be a timestamp or whatever instead, though. Users can intuitively understand that the final snapshot/branch is a precise reflection of the state of the logical database (not the physical database) as the time that the branching function returns, which maps to an LSN which is pretty much current the instant the function returns. That seems harder if snapshots have to physically serialize everything in-place (or get pageservers/walkeepers to do so synchronously). It seems easier if the implementation has more degrees of freedom at every layer.

I agree that if this is taken too far, you end up with something that is a snapshot in name only, which is bad -- no question. At the same time, materializing every page just to snapshot might well be *way* too pessimistic, and seems to work against the goal of deferring work that can be done in the background. Or work that can be avoided entirely in practice, through lazy processing. It will always work this way for a significant subset of all users -- users that are not going to be happy if the system has very significant write amplification for their use-case (at any tier of the system). I don't think that we can tolerate any truly pathological performance issues for even halfway-reasonable use cases.

PR #10, opened 2021-07-30
Yet another refinement of the how snapshots and WAL are managed

hlinnaka:
Based on what we discussed in offsite on Wednesday. This describes how we want things to work, not the current status quo.

### DIFF
--- a/retention-policy.md
+++ b/retention-policy.md
@@ -1,0 +1,91 @@
+# User-visible timeline history
+
+The user can specify a retention policy. The retention policy is
+presented to the user as a PITR period and snapshots. The PITR period
+is the amount of recent history that needs to be retained, as minutes,
+hours, or days. Within that period, you can create a branch or
+snapshot at any point in time, open a compute node, and start running
+queries. Internally, a PITR period is represented as a range of LSNs
+
+The user can also create snapshots. A snapshot is a point in time,
+internally represented by an LSN. The user gives the snapshot a name.
+
+The user can also specify an interval, at which the system creates
+snapshots automatically. For example, create a snapshot every night at
+2 AM. After some user-specified time, old automatically created
+snapshots are removed.
+
+                     Snapshot       Snapshot
+         PITR        "Monday"       "Tuesday"        PITR
+    ----######----------+-------------+-------------######>
+
+If there are multiple branches, you can specify different policies or
+different branches.
+
+The PITR period and user-visible snapshots together define the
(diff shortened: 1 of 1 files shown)


hlinnaka:
See also Ann's notes of the session: https://www.notion.so/zenithtech/Meeting-notes-on-28-07-2021-Storage-format-for-snapshots-e08ee6afc26649bbb1a287ee240f69a2

petergeoghegan:
@hlinnaka Are you saying that, in general, branches and snapshots neither require nor force any particular physical representation in underlying storage, before or after being created/restored?

To ask almost the same question another way: is there any sense in which a branch or a snapshot (or anything else) is at least roughly like a standard Postgres base backup? Are both branches and snapshots ultimately just logical concepts that don't strictly require any particular correct physical representation?

I think that I'm looking for a high level statement like:

Ultimately whether or not we retain WAL and base images is dictated by the retainment period (i.e. user requirements), as well as whatever the garbage collection process thinks will work best. Garbage collection is driven by back pressure. So we could use any mixture of extremely old and extremely new pages, depending on workload characteristics. Garbage collection continually makes trade-offs that weigh the need to bound tail latency from incremental per-page recovery against the need to avoid write amplification in S3. We want to save up at least a few WAL records per page for pages that are seldom modified.

Does that make sense?

hlinnaka:
On 31 July 2021 18:06:22 GMT+03:00, Peter Geoghegan ***@***.***> wrote:
***@***.*** Are you saying that, in general, branches and snapshots neither require nor force any particular physical representation in underlying storage, before or after being created/restored?

Yeah.

>To ask almost the same question another way: is there any sense in which a branch or a snapshot (or anything else) is at least roughly like a standard Postgres base backup? Are both branches and snapshots ultimately just logical concepts that don't strictly require any particular correct physical representation?

Right, they are both just logical concepts.

>I think that I'm looking for a high level statement like:
>
>Ultimately whether or not we retain WAL and base images is dictated by the retainment period (i.e. user requirements), as well as whatever the garbage collection process thinks will work best. Garbage collection is driven by back pressure. So we could use any mixture of extremely old and extremely new pages, depending on workload characteristics. Garbage collection continually makes trade-offs that weigh the need to bound tail latency from incremental per-page recovery against the need to avoid write amplification in S3. We want to save up at least a few WAL records per page for pages that are seldom modified.
>
>Does that make sense?

Yep!

PR #11, opened 2021-10-08
List of background tasks in page server.

hlinnaka:
The page server has several different background tasks it needs to or can perform on the layers. I've tried to list them here, based on the recent discussion on the discord channel.

This description and the terms don't match the actual code on 'main' currently. On 'main', the checkpointer thread performs 'Eviction' and 'Materialization' tasks together. The GC thread perfoms 'Garbage Collection', and 'Coalescing' and 'Bundling' are not performed at all.

### DIFF
--- a/background-tasks.md
+++ b/background-tasks.md
@@ -1,0 +1,38 @@
+# Eviction
+
+ Write out in-memory layer to disk, into a delta layer.
+
+- To release memory
+- To make it possible to advance disk_consistent_lsn and allow the WAL
+  service to release some WAL.
+
+- Triggered if we are short on memory
+- Or if the oldest in-memory layer is so old that it's holding back
+  the WAL service from removing old WAL
+
+# Materialization
+
+Create a new image layer of a segment, by performing WAL redo
+
+- To reduce the amount of WAL that needs to be replayed on a GetPage request.
+- To allow garbage collection of old layers
+
+- Triggered by distance to last full image of a page
+
+# Coalescing
+
+Replace N consecutive layers of a segment with one larger layer.
+
+- To reduce the number of small files that needs to be uploaded to S3
+
+
+# Bundling
+
+Zip together multiple small files belonging to different segments.
+
+- To reduce the number of small files that needs to be uploaded to S3
+
+
+# Garbage collection
+
+Remove a layer that's older than the GC horizon, and isn't needed anymore.


knizhnik:
The list of background tasks is determined by underlying storage model.
I.e., if we do not have in-memory layer, we do not need to evict something to the disk.
I agree that current layered storage requires this 4 tasks. Not sure if all of them should be done separately, by it own thread. It seems to be natural to combine  'Coalescing' and 'Bundling'  and perform them during S3 upload.

But I still  not quite convinced with layered storage model. So the unit of exchanging data between compute and page server is page. It is also not an axiom, for implementing bulk loading it may be ore efficient to use larger chunks. But let's first use this assumption. So we need map with page address as a key. Is it possible to hold this map in memory? Even if we use two-dimensional map <tblsp,dbnode,relnode>->relation and <pageid>->page, 
then we need at least 8 bytes per page. Can we accept that we can hold this map in memory? Please notice that I have not mentioned that this map has to be versioned and we need to include LSN as a key.

So, if we can not hold page map in memory, there are two alternatives:
1. Address larger units (segments or layers)
2. Use persistent map with some buffer cache.

From my point of view second approach is simpler, more flexible and reliable. Even if we failed to find or some existed persistent key-value storage for Rust or do not want to rely on third-party code, I think 
that straightforward implementation B-Tree with simple buffer cache is not so complex. And IMHO it is simpler than trying to implement layered storage. The main disadvantage of layered storage from my point of view is that it is difficult to find some efficient universal approach for splitting data into layers. Depending on user's data model and access pattern we can always  get touch number of too small files. Working with large number of files is always a challenge and  we rely on OS that it can efficiently handle thousands of files. This is Postgres approach, but I always think that database should be able to work with raw device (and most commercial DBMS except Postgres can do it).

If we have persistent map (KV storage), then there is no need in explicit eviction. Also 'Coalescing' is combined with 'Bundling' : S3 uploader just traverse KV storage and creates large fixed size segments for uploading to S3.

hlinnaka:
> The main disadvantage of layered storage from my point of view is that it is difficult to find some efficient universal approach for splitting data into layers. Depending on user's data model and access pattern we can always get touch number of too small files. Working with large number of files is always a challenge and we rely on OS that it can efficiently handle thousands of files. This is Postgres approach, but I always think that database should be able to work with raw device (and most commercial DBMS except Postgres can do it).
> 
> If we have persistent map (KV storage), then there is no need in explicit eviction. Also 'Coalescing' is combined with 'Bundling' : S3 uploader just traverse KV storage and creates large fixed size segments for uploading to S3.

The thing is, we will face this same "how to split data into layers" problem no matter what. Even if we have a persistent KV store to hold the page versions in the page server, we *still* have to decide how to split them into those fixed size segments for uploading to S3. You have the exact same tradeoffs there.

hlinnaka:
>  I agree that current layered storage requires this 4 tasks. Not sure if all of them should be done separately, by it own thread.

Yeah, I'm not sure how to schedule and trigger these. Currently, I'm imagining a thread pool, and a list of jobs with priorities.

> It seems to be natural to combine 'Coalescing' and 'Bundling' and perform them during S3 upload.

Agreed.

MMeent:
I think the 'coalescing' and 'garbage collection' problems can be solved by using a log-structured merge (LSM) tree. At ingest, we have raw wal. This is then moved into in-memory LSM layers for Segments / on-disk LSM layers for Relations / S3 LSM layers for the rest.

I think it is important to realize that we have 2 things we want to maintain: 

1. Latest page images for performant lookups of correct page versions
2. WAL logs for pages at specific versions.

A suggested LSM-inspired architecture:

- Layers exist of 2 components: (Sparse?) materialized data ranges at some LSN (let's call it the MatDatRange of the Layer), and the WAL streams of those MatDatRange.
- We maintain Layers at different levels of granularity: 
  - Segments, a 10MB stretch of Relation data (1280 8kB Pages, maybe move to 1024 for performance?), using BLOCK as it's materialized data
  - Extents of 128 (?) sequential Segments, using Segments as it's materialized data
  - Relations, using Extents as materialized data
  - Databases (maybe?), writing out Relation blocks.

- Each Segment's Layer maintains which blocks it contains, and how much WAL data it contains for that block.
- When a segment goes over THRESHOLD WAL data since last block image (not counting WAL overhead such as headers) for one block (e.g. 4x BLCKSZ), this block is imaged and added to this layer to maintain performance (NOTE: not registered as WAL)
- When a SegmentLayer is "full" (e.g. sizeof(WAL) + sizeof(latest pageversions) > 10MB), it is closed and a new Layer is created for that Segment.
- Full SegmentLayers are written to larger overarching layers, e.g. one Layer per PG Extent (1GB, ~ 100 Segments) or 1k Segments. 

- This ExtentLayer would contain compressed SegmentLayers: The latest PageVersion, and the WAL records (where not GC-able).
   - NOTE: We could make exceptions for this for Backups: retain the latest PageVersion+LSN for snapshots we need to keep.
- For each ExtentLayer we maintain which SegmentLayers are contained in this ExtentLayer for lookup performance.
- When an ExtentLayer is full (e.g. > 1 GB), we merge it into the RelationLayer log: Retain only interesting page images, and concatenate all WAL contained that cannot be dropped yet.

... etc

Note: A Layer can store it's PageImages and WAL streams seperately, for improved GC performance.

Benefits of this design are: 
- We always have recent versions of pages available without too much WAL replay.
- We retain wal seperately from latest page versions -> dropping WAL has never been easier
- Hot pages are generally always in memory
- ... others?

Potentially negative effects:
- We might have trouble finding that one old page in this hot Segment
  - Might be solved by having metadata for Layer that stores the latest LSN for each MatDRange that is not stored in this Layer.
- We could be storing hot pages many times in memory (at different versions, sure, but still)
- ... others?

knizhnik:
> The thing is, we will face this same "how to split data into layers" problem no matter what. Even if we have a persistent KV store to hold the page versions in the page server, we still have to decide how to split them into those fixed size segments for uploading to S3. You have the exact same tradeoffs there.

I do not see big problem here: just select records ordered by LSN starting from some point and pack them into fixed size segments. We can have special index by LSN, or use existed <buftag,lsn> index and just jump to the page using index lookup.

As far as I remember, we decided to keep all data (within PITR interval) at pageserver. It means that page server will never need to extract some particular page or range of pages from S3. Just in the case when it is crashed or we want to start new page server we may need to load most recent snapshot from S3. In this case this 1Gb segment files ordered by LSN are optimal. As Eric originally suggested we can have some incremental layers (i.e. only images updated since last checkpoint) interleaved with full layers (all images). Segments with WAL  are stored separately.

hlinnaka:
> > The thing is, we will face this same "how to split data into layers" problem no matter what. Even if we have a persistent KV store to hold the page versions in the page server, we still have to decide how to split them into those fixed size segments for uploading to S3. You have the exact same tradeoffs there.
> 
> I do not see big problem here: just select records ordered by LSN starting from some point and pack them into fixed size segments. We can have special index by LSN, or use existed <buftag,lsn> index and just jump to the page using index lookup.

And when do you materialize the pages, to reduce the amount of WAL that needs to be replayed on a GetPage request? And how do you release the space taken (in S3) by old page versions that are no longer needed? Exact same questions.

> As far as I remember, we decided to keep all data (within PITR interval) at pageserver. It means that page server will never need to extract some particular page or range of pages from S3. Just in the case when it is crashed or we want to start new page server we may need to load most recent snapshot from S3. In this case this 1Gb segment files ordered by LSN are optimal. As Eric originally suggested we can have some incremental layers (i.e. only images updated since last checkpoint) interleaved with full layers (all images). Segments with WAL are stored separately.

Again, you have exactly the same questions with those incremental layers or snapshots. When do you materialize the incremental layers into a new full layer? At what granularity?

These questions are very fundamental with any log-structured system like what we're building. We will have to tackle them head on.

knizhnik:
> I think the 'coalescing' and 'garbage collection' problems can be solved by using a log-structured merge (LSM) tree. At ingest, we have raw wal. This is then moved into in-memory LSM layers for Segments / on-disk LSM layers for Relations / S3 LSM layers for the rest.

IMHO it seems to be too complex.
I once again return to the simplest and at the same time most flexible solution, similar with what we already have with RocksDB:
just <tag,LSN>->[wal_record|page_image]
persistent key value storage. Not LSM, but normal B-Tree to avoid need in GC and merges.

For S3 we have separate data structure: large (~1Gb) segments of two kind:
1. Incremental snapshot  (page images changed since last snapshot)
2. Full snapshot (all page images at the given LSN)

Page store is not sending WAL records to S3. It is done by safekeepers (as original 16Mb Postgres WAL files or may be grouping them into larger segments).

When we want to start/recover page server,  we download from S3 last full snapshot and apply following incremental snapshots. Then we download WAL segments for the requested PiTR increval and scatter them.
It can be done using the same code which receives WAL from safekeeper.
Pros:
+ It is simple
+ Small number of large files
+ We are not storing WAL twice in original and scattered form

Cons:
1. Applying incremental scanpshots on top of full snapshots assumes extra network traffic.'
2. The same is true for extra storage overhead: if some pages is updated multiple times, it also will be included in many incremental snapshots.
3. Scattering WAL can take some extra time (but comparable with time, but it seems to be much smaller than time needed for download this WAL from S3).

Please notice that we rarely will need to start new or recover old page server.
In most cases we just stream data to S3 and do not read it. As Eric told, having quite simple file structure at S3 (just segments with Postgers pages) it is possible to restore Postgres database using very simple utility.
Right now decoding our layers is not possible without out Rust code. In principle, we can make this structure even compatible with probackup or some other backup tools supporting incremental backups. But I am not sure that it is needed.

knizhnik:
> And when do you materialize the pages, to reduce the amount of WAL that needs to be replayed on a GetPage request? And how do you release the space taken (in S3) by old page versions that are no longer needed? Exact same questions.

Page are materiaized:
1. During GC (to be able to cut old history)
2. get_page_at_lsn (on demand)
3. By S3 uploader (it materialize all pages updated since last checkpoint, by checkpoint here I means upload to S3).

Once we create full snapshot, we can remove all previous incremental snapshots (if we do not need them for PITR)

knizhnik:
> Again, you have exactly the same questions with those incremental layers or snapshots. When do you materialize the incremental layers into a new full layer? At what granularity?

This problem is many time solved by probackup and other incremental backup users.
The simplest criteria is to create new full snapshot when size of incremental snapshots exceeds size of full snapshot or when number of such incremental snapshots exceeds some threshold value

Yes, I know disadvantages of this approach. If just one page of the database is frequently updated, then we will create incremental snapshots with just one page. As a result we will have very large number of incremental snapshots.  It will increase time needed to restore page server.  And when we create new full snapshot, we have to waste time and network traffic copying data which was not changed. But IMHO it is corner case. If average size of our user's database is expected to me quite small (few gigabytes), then it doesn't matter at all: it is cheaper and faster to copy full  database rather than playing with small layers.

So as plan-0, granularity can be the whole database instance: so we just split the whole database in 1Gb segments and that is it. Please notice, that this segments doesn't match Postgres segment. Them can contain pages from different relations. It's size also can be larger or smaller than 1Gb (but fixed size).

knizhnik:
But it will be not difficult to add more granularity and efficiently handle really large databases, only small part of which is updated. What is the best way to split pages between this ??? (may be "areas")? I don't know. It can be Postgers database. Or just splitting relations between different areas by hash. None of this approaches is perfect and suffers from data skews. But this division into areas need not to be fixed and can be dynamically changed. What we want to reach is that pages are splitted between segments in more or less consistent way. And if most pages are not changed, than most segments are also not changed. In this case there is no need to upload such segments to S3 once again.

MMeent:
> IMHO it seems to be too complex. I once again return to the simplest and at the same time most flexible solution, similar with what we already have with RocksDB: just <tag,LSN>->[wal_record|page_image] persistent key value storage. Not LSM, but normal B-Tree to avoid need in GC and merges.

We need to garbage collect full page images and stored WAL from the pageserver to clean up older stuff from our systems and save on resources. How would you apply GC to your suggested 'just <tag, LSN>->[wal_record|page_image] persistent key value storage'?

> For S3 we have separate data structure: large (~1Gb) segments of two kind:
> 
> 1. Incremental snapshot  (page images changed since last snapshot)

When is an FPI included in this incremental snapshot? When it has one change? Does this incremental snapshot somehow not contain WAL? Should we maintain extra metadata?

> 2. Full snapshot (all page images at the given LSN)

Is this for a subset of pages in a range, or one continuous Segment (?) of pages at the same LSN?
 
> Page store is not sending WAL records to S3. It is done by safekeepers (as original 16Mb Postgres WAL files or may be grouping them into larger segments).

Grouping WAL records by page / Segment is probably a good idea, as it allows for more data locality when recovering specific pages. That in turn then allows for faster full-cluster recovery due to better data locality, or point data recovery due to less seeking required.

> When we want to start/recover page server, we download from S3 last full snapshot and apply following incremental snapshots. Then we download WAL segments for the requested PiTR increval and scatter them. It can be done using the same code which receives WAL from safekeeper. Pros:
> 
> * ...
> 
> * We are not storing WAL twice in original and scattered form

What does the incremental snapshot contain, if not WAL?

knizhnik:
> We need to garbage collect full page images and stored WAL from the pageserver to clean up older stuff from our systems and save on resources. How would you apply GC to your suggested 'just <tag, LSN>->[wal_record|page_image] persistent key value storage'?

In the same way as it was implemented before: just remove everything beyond PiTR interval and materialize page at cut-off LSN if needed. B-Tree can quite efficiently implement range removes so I do not see big problems with fragmentation and whatever else here.

> When is an FPI included in this incremental snapshot? When it has one change? Does this incremental snapshot somehow not contain WAL? Should we maintain extra metadata?

My suggestion is very simple. There is checkpointer (or uploader process) which is triggered by time or amount of written WAL. It just traverse KV storage and materialize  pages at snapshot LSN (if them were not materialized) before.This pages are appended to segments which are sent to S3. Each segment is fixed size (few Gb) and contains pages from different relations. Segment metadata (page addresses) is stored locally in separate file, so that it is possible to inspect segment content without downloading segment itself.

The difference between full and incremental backup is only that in first case we materialize pages which were updated single last snapshot LSN, and for full backup - all pages.  Certainly only if them were not materialized before.

knizhnik:
> Is this for a subset of pages in a range, or one continuous Segment (?) of pages at the same LSN?

Pages are sorted by tag because we are traversing them in this way.  Each segment contains fixed number of pages from different relations. As far as we never need to extract some particular page from S3, we do not need some efficient mechanism to lookup page at S3 by tag. So we just download segment with its metadata and insert them in KV storage. But once again - it will rarely happen. In most cases we just send data in one direction. We may need to download metadate of segment from last full backup to check that new segments are identical to old ones. This is why I prefer to store metadata independently -to make it possible to load it without loading data itself.

knizhnik:
> Grouping WAL records by page / Segment is probably a good idea, as it allows for more data locality when recovering specific pages. That in turn then allows for faster full-cluster recovery due to better data locality, or point data recovery due to less seeking required.

May be I missed something. Recovery of pages is done by pageserver, right? And it is not swapping its data to S3 (unlike safekeeper). It stores all its data which may be needed for recovery locally.
So we never need to request from S3 WAL  to reconstruct some particular page.

Please correct me if I am wrong, but my understanding at this moment is that we need data from S3 only if page server local storage is corrupted and we have to reload its storage from S3. There is no some way to detect which particular part of data is corrupted or missed, so there is no other way rather than to reload all data. But fortunately it will happen very rarely.

Also we may need to launch new page server to perform load balancing. May be even in different geographical zone. But in this case it is also necessary to download all data from S3.

knizhnik:
> What does the incremental snapshot contain, if not WAL?

Pages changed since last snapshot.

MMeent:
> May be I missed something. Recovery of pages is done by pageserver, right? And it is not swapping its data to S3 (unlike safekeeper). It stores all its data which may be needed for recovery locally.

If I wanted to do PITR (e.g. create a timeline sometime in the past where we can still guarantee the data exists but exists only in Incrementa/Full Snapshots, +wal) then I think it is very reasonable to pull that data from S3 to recover. And in such situations, it should be relatively straightforward to pull and re-apply WAL if we've decided to store the applied WAL right next to the incremental snapshot that was created from that WAL.

> So we never need to request from S3 WAL to reconstruct some particular page.

Unless we flushed this WAL for GC due to resource contention. I really think we can support PITR windows way beyond what WAL we keep available on the PageServer IF we want to, as long as the WAL is available in an efficient-enough format. 

Similarly, Read-only nodes need a consistent view of the whole database. When they lag behind due to a long-running read-only transaction, they might need to access now-GC-ed data to maintain a consistent snapshot.

knizhnik:
> If I wanted to do PITR (e.g. create a timeline sometime in the past where we can still guarantee the data exists but exists only in Incrementa/Full Snapshots, +wal) then I think it is very reasonable to pull that data from S3 to recover. And in such situations, it should be relatively straightforward to pull and re-apply WAL if we've decided to store the applied WAL right next to the incremental snapshot that was created from that WAL.

I do not believe that somebody wants to do PITR at the arbitrary point long in the past, outside GC interval which we can keep at page server. Also it seems to me that users most likely will want to rollback to some predefined snapshot (i.e. before performing some dangerous operation we created snapshot so that we can undo it). Such snapshots can be naturally integrated with our checkpoints.

And last my argument: out measurements shows that scattering WAL is not a bottleneck.  If we really need to receive missed WAL at pageserver, we can request it from safekeeper (which in turn can download missed WAL segments from  S3) and time of scattering/storing WAL will be negligible, comparing with overhead of retriving this WAL from S3.

So I do not think we actually need to requst scattered WAL to pageserver from S3. And so there is no any reason to have the same format of data for layered storage and S3. Please notice, that even in your proposal presence of 
 'Coalescing' and 'Bundling'  steps makes this formats not compatible. It seems to be oblivious that we can't send to S3 layer file as it is (because them are too small for S3), so we have to perform  'Coalescing' and 'Bundling'.
But we also can not use file received from S3 as it is in layered storage, because deserializing it will be too expensive and consume a lot of memory. So in any case we have to make reverse transformation and split S3 segment into several layers.  So there will be not o much difference from just reinserting page images from received S3 segment in KV storage.

MMeent:
You've convinced me that for recovery from S3 this is not an issue. So, your proposed solution for non-local storage is:

S3 Layers: data by the GBs. E.g. by PG Extent ->1GiB continous data
- Differential snapshot: 
    Sparsely populated; only page images of the pages changed since last Full Snapshot in the NBLCKS range.
- Full Snapshot: 
    A continuous range of NBLCKS blocks of a relation (Segment? Area? Extent?) @ LSN.
    Generated when the size of differential snapshots fo this Segment/Area/Extent exceeds threshold

In general, I agree with this. We'd need to add some metadata (see below) for indexing and S3 layer GC, but on the data part I agree.

I'd like to ask, however, what the plan is with local data. 

I think that storing everything in one big store is very expensive for unchanged data; each page and WAL record would need metadata on the levels of the full page identifier (LSN (8) + PageId (4) + Forknum (1) + RelId (4) + DbId (4) + Timeline(16) + Tenant (16) = 53 bytes). For pages, that is only 0.6% overhead, but for WAL records (which I suspect will generally be much smaller than one page) that overhead will be much more significant.

---

Each S3 Layer should probably have metadata about:
- Which Area/Segment/Extent it contains
  - useful for debugging & error detection.
- When blocky:
  - Which of the blocks are contained in this Layer
     So that we can know if this layer is interesting, without having to scan through the actual data.
  - What is the biggest LSN for each Block in this Area/Segment/Extent (even if no data is stored for that Block in this Layer)
    This is mainly so that we can immediately skip to older Layers if we need to restore some old block that hasn't changed for 30 incremental snapshots, without having to touch the layers in between.
    


PS. I find it slightly ironic that you suggest using RocksDB instead of a specialized LSM tree, as RocksDB's storage format is an LSM tree

knizhnik:
Sorry, I am not suggesting RocksDB.
I think that we need to have persistent KV storage, but it should not be LSM.
What I am currently have in mind is standard B+Tree be fixed size keys and values.
Value is just reference to BLOB storage. BLOB storage is consists of two parts:

1. SLRU storage for WAL record consistent of large fixed size  segments (files). All received WAL records are appended to this storage. And old segments outside PITR interval are deleted.
2. Page storage. It is just one large files with fixed size pages. Once page is materialized by wal-redo, it is placed in this storage. When page version is not needed any more, it is deleted, but linking in list of free pages.

So allocation/deallocation in this sortages requires o(1). No caching is needed for them: data is read/written directly to the file. B+Tree use simple buffer cache and fixed size (page) allocator. It is not quite clear to privide fault tolerance. Use WAL, maintain shadow pages...

As you can see we never have to copy/rewrite large bulks of data.
The key point is that format of storing data in KV storage and  S3 are completely independent. So we can change one of them without affecting another. When we create files for S3, we just traverse KV storage, extract and copy data from it to S3 segments. And when we import data from S3, we just reinsert pairs in KV storage.

knizhnik:
> I think that storing everything in one big store is very expensive for unchanged data; each page and WAL record would need metadata on the levels of the full page identifier (LSN (8) + PageId (4) + Forknum (1) + RelId (4) + DbId (4) + Timeline(16) + Tenant (16) = 53 bytes). For pages, that is only 0.6% overhead, but for WAL records (which I suspect will generally be much smaller than one page) that overhead will be much more significant.

We can have hierarchy of B-Trees: 
```
    tenant_id+timeline_id (32 bytes)
          tblspnode+dbnode+relnode+forknum (13 bytes)
                block_no+block_offs+LSN (14 bytes)
```
Actually top B-Tree may be not needed:  we can put each timeline in separate file.

MMeent:
> block_no+block_offs+LSN 

Which block_offs is this about? Our units of storage are 8kiB (well, BLCKSZ) of data, and WAL records, right? And postgresql has a maximum of u32 blocks /relation, regardless of RELSEG_SIZE.

> Sorry, I am not suggesting RocksDB.

OK, sorry for misunderstanding

> I think that we need to have persistent KV storage, but it should not be LSM.
> What I am currently have in mind is standard B+Tree be fixed size keys and values.
> Value is just reference to BLOB storage. BLOB storage is consists of two parts:
> 1. SLRU storage for WAL record consistent of large fixed size segments (files). All received WAL records are appended to this storage. And old segments outside PITR interval are deleted.
> 2. Page storage. It is just one large files with fixed size pages. Once page is materialized by wal-redo, it is placed in this storage. When page version is not needed any more, it is deleted, but linking in list of free pages.
> So allocation/deallocation in this sortages requires o(1). 

Sure. But this would mean that scanning a relation sequentially at PG side (with empty buffers) would then do a random walk across the BLOB of page versions, which is quite a bit more expensive than a normal sequential scan; which we would need to tune PG for. Especially if we are using some non-node-local data store (like EBS). 

Q: Do we rely on ephemeral storage nodes such as r5d for low-latency local SSDs?

knizhnik:
> Which block_offs is this about? 

Sorry, looks like I thought about something else:(
Certainly it is enough to have block_no, so size of  key will be 4+8=12 bytes.
So does it make sense to reduce key size twice by splitting single tree in a "forest"? It was my original intention, but then I found out that it leads to significant complication of KV storage(both API and implementation).
So may be it was not so good idea. We can use prefix compression in single tree to reduce key space overhead.

> Sure. But this would mean that scanning a relation sequentially at PG side (with empty buffers) would then do a random walk across the BLOB of page versions, which is quite a bit more expensive than a normal sequential scan; which we would need to tune PG for. Especially if we are using some non-node-local data store (like EBS).

But subsequent pages are not be always  located sequentially in the current layered storage as well.
When checkpointer materializes delta layer, it reconstructs only those pages which were updated, doesn't it?
It means, that if relation is not read-only and we randomly update its pages, then we also do not have image layers containing all pages of relation.

Also, it seems to me that problem with seqscan is mostly related with get_page_at_lsn round-trip, so it should be first of addressed by prefetch/prewarm. But certainly possibility to fast read of sequential pages by pageserver is also important.

> Q: Do we rely on ephemeral storage nodes such as r5d for low-latency local SSDs?

I think Zenith will not work normally without fast local storage. Especially on OLTP workloads.

knizhnik:
Currently Postgres WAL record is 24 bytes, so if we reduce size of key from 25 bytes to 12 bytes then we save less than 30%. And it is for empty WAL  record. Size of most frequent HOT_UPDATE/INSERT records is greater than 80 bytes. For them difference will be less than 12%.

MMeent:
> > Sure. But this would mean that scanning a relation sequentially at PG side (with empty buffers) would then do a random walk across the BLOB of page versions, which is quite a bit more expensive than a normal sequential scan; which we would need to tune PG for. Especially if we are using some non-node-local data store (like EBS).
> 
> But subsequent pages are not be always located sequentially in the current layered storage as well. When checkpointer materializes delta layer, it reconstructs only those pages which were updated, doesn't it? It means, that if relation is not read-only and we randomly update its pages, then we also do not have image layers containing all pages of relation.

That is indeed the case right now; but it could (should?) be different in a changed design. 

E.g. Full layers are consecutive write-outs of 1GB extents of the table, Delta layers are sorted arrays of the changed blocks, and only blocks changed since last checkpoint are written into BLOBs.
The checkpointer then regularly writes out the randomly dispersed changed pages into new delta layers.

> Also, it seems to me that problem with seqscan is mostly related with get_page_at_lsn round-trip, so it should be first of addressed by prefetch/prewarm. But certainly possibility to fast read of sequential pages by pageserver is also important.

Yes. Thankfully, postgres already has some prefetching systems. I'm not certain we hook into that, though...

> > Q: Do we rely on ephemeral storage nodes such as r5d for low-latency local SSDs?
> 
> I think Zenith will not work normally without fast local storage. Especially on OLTP workloads.

EBS is generally fast enough for same-node accesses, but we're introducing our own latency through MITM-ing postgres' IO ops. I think we should generally be very sensitive to latency and prevent as many IO ops per `page@LSN` call at the pageserver as we realistically can.

MMeent:
Regarding which tasks: I see some 6 tasks in the pageserver, of which 2 (3) in the primary flow, and 4 (3) background. See also: [0]

1. Sending pages to Compute (FG)
  a. if 3 is not up-to-date, apply WAL where applicable
2. Receiving WAL from safekeeper (FG)
3. Replaying received WAL (BG, but can be called from 1)
4. Writing out differential layers (BG)
  a. uploading these layers
5. Compacting layers (BG)
  a. uploading these compacted layers
6. Garbage collection (BG)


[0] https://www.notion.so/zenithtech/Matthias-sketch-book-on-PageStore-and-other-storage-related-scribbles-12b877b71945416eaff3cf4cf8b2f970#c11f2bb437284583a4e947d1d3047228

PR #14, opened 2022-01-18
RFCs: Storage; structures and new layer format

MMeent:
- [Storage and file formats; Structures](https://github.com/zenithdb/rfcs/blob/storage_and_layers/text/layer-physical-formats.md) (informational)
- [Differential Layer Format](https://github.com/zenithdb/rfcs/blob/storage_and_layers/text/differential-layer-format.md)


text/layer-physical-formats.md
```diff
@@ -0,0 +1,87 @@
...
+When we *do* eventually find some of the data we need to restore the requested page version, then that data should be
+complete enough to restore the full page; i.e. for some WAL / delta records found in one file, we should not have to
+go to another file to request the base page version. This implies that each page with deltas in a delta file will also
+have a base image in that same delta file. *This indeed wastes some space (with a bad pathological ~ 400x worst case of
+a single minimal WAL record on that page in this layer), but this is important in the long run; as it saves us from
+cross-file dependencies.*
+
+### On disk space savings methods: Physical representation of page version metadata
+
+As Branches are always replayed in ascending order of LSN, we should store the WAL in ascending order as well; so that
```
knizhnik: If *Branch* is page image + deltas till next page image, then I do not understand what you mean by "Branches are always replayed in ascending order of LSN".  `getpage@LSN` needs only one page, so it needs to replay only one *branch*. And definitely all WAL records in the *branch* are stored in LSN ascending order.

text/differential-layer-format.md
```diff
@@ -0,0 +1,143 @@
...
+store some metadata in the BranchPtr which discerns the kind of branch that
+pointer points to. This allows for compacter storage of the actual branches,
+at the cost of some extra data in the
+
+Not all branches will describe multiple page versions: It is indeed possible
+that a page will be updated only once during a checkpoint.
+
+In that case, it
+is quite wasteful to store the metadata of '0 page versions'
+
```
knizhnik: Both in-memory layer and delta layer now contains data only of changed pages.

text/differential-layer-format.md
```diff
@@ -0,0 +1,143 @@
...
+Additionally, because the length of the WAL record is stored in it's header, we
+don't actually need to store the length of the WAL record in the branch structure;
+we can read it directly from the header, saving 4 bytes per WAL record.
+
+With LSN array compression and the emission of the WAL record length also being
+stored outside the WAL record, we can reduce the storage overhead of non-FPI WAL
+records from 28 bytes per record down to less than 8 bytes per record, with an
+expected value of ~ 4 bytes (depending on how often a page is modified; more
+frequent / less noizy WAL record generation = lower overhead).
+
```
knizhnik: I partly agree with your arithmetic. But...
Right now delta_layer is storing 4 byte length + serialized PageVersion.
PageVersion itself is itself nested enum, with 8 bytes used just for enum tags (this seems to be the first optimization target, because one byte is enough to encode it). Size of Postgres XLogRecord header is 24 bytes.
Length is really stored inside XLogRecord. Typical size of record (i.e. update of integer field is 52 bytes - just check WAL log for pgbench).

So what we have. Right now we are definitely wasting 7 bytes just because of serialization, we can remove extra 4 bytes with record length. It will complicate traversal (to be able to deserialize data we first need to extract it and to know how much data we need to read, we need first to extract length). But in principle it is possible.
Using delta encoding we can encode LSN using just few bits. So we can throw away 20 bytes.
So instead of 12+52=74 bytes we will store 54 bytes. So we save about 30%. It is certainly  noticeable, but ... do not radically change storage and memory demands.

text/differential-layer-format.md
```diff
@@ -0,0 +1,143 @@
...
+### Page versions array
+The Delta layer uses an in-memory sorted array of <lsn, pageversion> pairs to
+represent the changes to a page. This is correct and allows for O(log N) version
+lookups (for N = number of total page versions). This is OK, but can be improved
+upon: as mentioned in "Storage and file formats; Structures", we always
+need a base page image to reconstruct a page. By limiting the array of page
+versions stored for each page in the Layer to Branches, we eliminate a
+significant amount of entries from the search space; improving this search speed
+by O(log fraction of non-image page-versions) and reducing memory usage.
+
```
knizhnik: Right now delta layer stores only WAL data. Reconstructed pages are stored in image layer.
Delta layer may contains page images only if WAL contains FPI-s. Only in this case there can be multiple *branches* in inmem/delta layer. I do not understand what do you suggest to change in this schema.

text/differential-layer-format.md
```diff
@@ -0,0 +1,143 @@
...
+      Lsn -> BranchPtr -> ...
+      Lsn -> BranchPtr -> ...
+      ...
+```
+
+Note that the definition of BranchPtr is partially dependent on what data we
+want to store in the BranchPtr, and what data we want to store locally in the
+Branch. Either way, the overhead of one branch in the page versions array is
+16 bytes.
+
```
knizhnik: What is the intended accessed pattern?
The obvious assumption is that compute node is not requesting recently updated pages.
So page server has some time to reconstruct them in background. It is now done by checkpointer,
but it may be better performed by separate thread to distinct persisting data and page reconstruction.
In the ideal case, page is reconstructed be checkpointer (reconstructor) before it is requested by get_page_at_lsn.
In this case latency of get_page_at_lsn will be minimal. And we do not need random access at all!
Checkpointer can sequentially read all data from delta layer and reconstruct page images. So there is no need to use some more sophisticated data format to provide to the branch. 

The main question is what to do with reconstructed images? Storing them in image layer may cause write amplification. If them are stored in page cache, then eviction algorithm will make part of this reconstuction work useless, because reconstructed page is evicted before it is accessed. I do not have good answer for this question.
May be we can really do not perform eager page materialization and always do it only on demand. But in this case reuestion page from pageserver will never be fast enough.

text/differential-layer-format.md
```diff
@@ -0,0 +1,143 @@
...
+larger segments means larger amounts of page versions in a layer, which means
+larger lookup times.  
+In the Delta layer there is also no description on when a page was last modified,
+so any failed lookup in the Delta layer must search back to the previous layer,
+making such lookups in the worst case O(log N * M) for N pageversions per layer
+and M layers. This currently is not too much of an issue, because this M is
+limited to 1 due to how our layers work: we always have an image layer
+immediately before the delta layer, which limits the number of Delta layers we
+access to 1.
+
```
knizhnik: It will be nt true if we commit PR https://github.com/zenithdb/zenith/pull/1133

text/differential-layer-format.md
```diff
@@ -0,0 +1,143 @@
...
+### Segment size
+The segment size of the current layer formats is very small; which means that
+the number of layers is relatively high. As each layer has some constant
+overhead, it should be beneficial to increase layer size to something in the
+order of gigabytes, instead of megabytes.
+
+The Differential layer format is designed for segment sizes of up to 8GiB,
+almost 1000 times larger than the current segment size. 
+
```
knizhnik: Write amplification for image layers is proportional to layer size. Certainly we can postpone generation of image layers. But not so long. And worst case scenarios may cause x1000 and even more write amplification for 1Gb segments.
Certainly we can use incremental image layers. But in this case get_page_at_lsn will have to read many files and perform searches in them.

text/layer-physical-formats.md
MMeent: A better phrasing could indeed be "As each branch is always replayed in ascending order of LSN, ...". 

Does that make the intention more clear?

text/differential-layer-format.md
MMeent: The access pattern would be either page@Lsn(latest), in which case we can use branch index 0;  or alternatively, the access is page@Lsn(something), in which case a simple binary search suffices (which in this case has a significantly smaller search space (presumably, orders of magnitude) than the current Delta layer).

text/differential-layer-format.md
MMeent: The code in main does not seem to have partial image layers.

text/differential-layer-format.md
MMeent: am I looking at the wrong branch? [DeltaLayerInner](https://github.com/zenithdb/zenith/blob/main/pageserver/src/layered_repository/delta_layer.rs#L148) most definately uses an VecMap<(SegmentBlk, Lsn), BlobRange> to store the references to versions. `SegmentBlk (4) + Lsn (8) + BlobRange (8 + 8) = 28`. This is not even counting the overheads of the PageVersion storage itself.

> It is certainly noticeable, but ... do not radically change storage and memory demands.
The main part is that this optimization of Branches is targetted to storage; as the data of the Branch itself will only reside in memory whenever the page is actually being requested.

Sequentially stored Branches allow us to store the location of required data much more concisely; decreasing memory usage significantly.

text/differential-layer-format.md
MMeent: That would expose the `O(M * log(N)` issue, yes.

text/differential-layer-format.md
MMeent: That is exactly why I implement a page index in each layer, which either points to the branches for that page in this layer, or contains the last LSN that this page was modified. No need for intermediate image layers; only one base image layer required (to allow us to remove older WAL). Updates can then be flushed down onto that one base image layer when the layer containing those updates are removed from the local disk.

text/differential-layer-format.md
MMeent: > WAL contains FPI-s

This is currently likely to happen, due to interplay of PG checkpointer, WAL-logging of hint bits, and the FPW being on by default.

> I do not understand what do you suggest to change in this schema.

What I suggest to change in this schema is to loosen the restriction to only allow WAL and WAL-FPIs in non-Image layers, so that we can store Branches instead, which are complete descriptors of the changes to a page over some period in time. This allows us to only store one pointer to that data in memory instead of the N pointers to the N changes to that page during that period.

text/layer-physical-formats.md
```diff
@@ -0,0 +1,87 @@
...
+At some point, storing more deltas is not worth the effort; it becomes clear that we’re not going to see requests from
+Compute with LSN < X. At that point, we could replay all completed segments <X onto existing image layers; so that we
+can drop the segments locally and save local storage space.
+
+### On Pageserver optimizations: write amplification / image layers
+
+When we have no need for intermediate image layers between layers which include
+deltas (due to e.g. low overhead of finding the right page version), we can
+decrease write amplification of creating and maintaining image layers by
+maintaining one "base" image layer, whose pages are modified and updated when
```
SomeoneToIgnore: > whose pages are modified and updated

From my (very limited) S3 remote storage standpoint, I'd love to see what changes in our checkpointing model and how that affects the fact that now no already created files are ever updated.

For now, before LSM trees (or whatever else) is implemented, we're bundling new files from checkpoints together into a streaming (zstd) archive and rely on this "never update" file approach when uncompressing, skipping the files existing locally (except the timeline metadata file).
Due to the same assumptions, we compress the checkpoint files without checking either, since they can add new files only.

So far, only this particular place seems worrying to me, but do you have any concerns?

MMeent: The main difference between the current (potential future) checkpoint code would be that
 1.) top-down checkpoints (initiated by influx of WAL) won't need to generate Image layers, and 
 2.) GC checkpoints would potentially update the bottom-most Image layer, then upload that layer with the new LSN attached. 

Note that the upload of the Image layer would not strictly be required as long as the intermediate layers still exist in S3; as we'd basically be 'inlining' the `base + gc_able_differences` in the `base_image_layer + gc_able_page_deltas + new_page_deltas = requested page version` formula.

SomeoneToIgnore: Right, so more logic, connected to GC.
We have nothing on that front in pageserver s3 sync now, but with the potential _updates_ we should adapt the s3 sync logic anyway, so I would love to see a "new GC and pageserver's local file implications" section if/when any further development happens, so that I could pick up those updates on my side.

(AFAIK, LSM approach is also updating some files on rebalancing? If so, we're on the right way with adjusting the existing code to such events)

text/layer-physical-formats.md
knizhnik: Yes

text/differential-layer-format.md
knizhnik: Sorry, I mean in-memory and delta layers

text/differential-layer-format.md
knizhnik: I looked mostly at in-memory layer layout.
I agree with you, 28 bytes is really too much overhead for in-memory part of delta layer.
Using RLE (run-length-encoding) and delta encoding we can really reduce 28 bytes to ~4 bytes.

text/differential-layer-format.md
knizhnik: PG checkpoints can be actually disarmed (by setting max_wal_size to vey large value), and wal logging of hint bits IMHO also should be switched off.

So you suggest to include page images in delta layer (as part of *branch*)?
Having just one pointer to brnach in deta-layer header definitely makes sense.

text/differential-layer-format.md
knizhnik: Yes

text/differential-layer-format.md
MMeent: >  Both in-memory layer and delta layer now contains data only of changed pages.

Yes. That is also, but partially, the case for the Differential layer format: it will not store actual page data, but will add an (technically optional, but very useful) last-changed-at-LSN for each page that has updated between some known LSN (e.g. the previous Image layer, or the PageServer's GC horizon) and the start LSN of this Layer.

The "0 page versions" is referring to the case where there's only one FPI (page image or record with will-init) being written for that page in a layer, without subsequent WAL records, through e.g. dirty bits. That means no other changes exist for that page, so we only have a 'head' in the _branch_, with the 'tail' of WAL to be applied to the page missing. As for the usual case of a _branch_ having 'tail' items, it would need to store the count of 'tail' items, but if we differentiate between the types of _branches_ we could treat _branches_ with and without 'tail' as different types of branches; removing the need to store the metadata that describes the 'tail' items.

text/differential-layer-format.md
MMeent: Yes. I'll note that I havent looked too much into the InMemory layer, as that's something that will likely change significantly when LSM is implemented. Although that too could make use of _branch_ optimizations, the implementation of those optimizations into InMemory layers is not in the scope of these RFCs.

text/differential-layer-format.md
MMeent: > So you suggest to include page images in delta layer (as part of branch)?

Yes, as that would reduce / remove the need for intermediate Image layers as a performance booster for the `pageimage<@LSN`-requests required to replay WAL records. I admit that that would introduce some write overhead; but with increased Segment and Layer sizes the extreme write amplification case of one page per record should occur significantly less often.

knizhnik:
Trying to implement checksums (CRC) for layer files I realized that it is hard to do with current layer file format and so it should be considered in new layer format design. 

The problem is that calculating CRC for the whole layer files may be very inefficient (especially checking it). And actually not needed:
WAL records are protected by CRC (stored in record header).
Page images will be also protected by checksums once them are enabled.

So we should protect by  CRC only "metadata" part of the layer. It is expected to be small and so CRC calculation should not take a lot of time. Unfortunately right now it is problematic. First of all because of use of bookfile which stores its own state.
Second - because of use serde serializer which packs together both "data" and "metadata".

So IMHO new layer format should separately store data (WAL records or page images) and metadata (everything else).
Mostly it means that we should not use standard serializers and bookfile.

One more argument for replacing bookfile is file compression. It is expected to be very useful because allows not only to save expensive local storage space, but also increase performance (because of reading smaller amount of data from the disk).
I have implemented prototype of compressed storage (each page is stored as bookfile chapter). It was quite simple, but inefficient. Each chapter is represented by hash table entry in a book. And loading file metadata takes a lot of time and consume large amount of memory. Instead of hash, we can just have array of page offsets. Memory footprint in this case should be much smaller.

MMeent:
> Trying to implement checksums (CRC) for layer files I realized that it is hard to do with current layer file format and so it should be considered in new layer format design.

Agreed

> The problem is that calculating CRC for the whole layer files may be very inefficient (especially checking it). 

Agreed, though a whole-file checksum is still very useful for up/download verification, and file integrity checks without needing to parse the file, in situations such as reboots or other situations where we would want to validate the local file system.

>  WAL records are protected by CRC (stored in record header).

This is only true for the `ZenithWalRecord::Postgres` variant of WAL records. Other variants do not (yet?) contain the metadata to determine record corruption.

> So IMHO new layer format should separately store data (WAL records or page images) and metadata (everything else). Mostly it means that we should not use standard serializers and bookfile.

Do you suggest storing these in different files? 
Would you consider the information contained in a 'branch' abstraction (i.e. the full data structure of the 'branch', incl. LSNs, records, base image, etc.) as "data" or as mixed data+metadata?

> One more argument for replacing bookfile is file compression

I generally agree; but I don't think that bookfile in itself is prohibitive for compression. The _branch_ abstraction allows for a compressed data sections for that branch, which should help quite a lot already.

> Instead of hash, we can just have array of page offsets. Memory footprint in this case should be much smaller.

Yes; I did not intend to implement the page index as a hash; but as a compressed bitmap. This should allow for low memory and storage overhead when page updates are randomly dispersed; while at the same time maintaining limited memory and compute overhead in all but the most abysmal of cases.

stepashka:
let's move the relevant part of this PR to zenith repo and close this

PR #15, opened 2022-01-19
Cluster size limits RFC

lubennikovaav:


text/cluster-size-limits.md
```diff
@@ -0,0 +1,74 @@
...
+First of all, it's necessary to define timeline size. Options are:
+
+1. Count all data, including SLRUs. (not including WAL)
+Here we think of it as of a physical disk underneath postgres cluster.
+This is how `LOGICAL_TIMELINE_SIZE` metric is implemented in pageserver right now and the draft PR uses it.
+
+2. Count only relation data. As in pg_database_size().
+
+This approach is more user-friendly, because it is the data that is really affected by user.
+To implement it, we will need to refactor timeline_size counter, or rather add another counter.
```
LizardWizzard, line 40: Currently LOGICAL TIMELINE SIZE doesn’t count everything, I remember that it doesn’t count TwoPhase files and as you mentioned wal is not counted too. So I think it can be just changed in place. Some time we will want to have a counter for data including wal. So I think two counters is enough

LizardWizzard (commented):
We treat compute as untrusted component, that’s why we try to isolate it with secure container runtime or a VM. In this RFC check for capacity is performed only on compute side, so I think this is vulnerable. Can user change this GUC by altering pg_setting (IIRC it is called this way)? If we allow installation of custom extensions this can really open the door. And of course some vulnerability may be used too. 

So I think we need some guard mechanism on the pageserver too.

lubennikovaav:
> We treat compute as untrusted component, that’s why we try to isolate it with secure container runtime or a VM. In this RFC check for capacity is performed only on compute side, so I think this is vulnerable. Can user change this GUC by altering pg_setting (IIRC it is called this way)? 

It depends on the GUC context. 
If we choose PGC_POSTMASTER context (any change requires restarting the server) it will be totally secure in our setup, but not flexible at all.
Another option is PGC_SIGHUP context ( Send a SIGHUP signal to the postmaster to cause it to re-read postgresql.conf and apply the changes). This allows us to change the limit on a running compute, which can be useful sometimes.
The user shouldn't be able to change it unless it has access to the VM.

If we allow installation of custom extensions this can really open the door. And of course some vulnerability may be used too.
> 
> So I think we need some guard mechanism on the pageserver too.

I agree that we need a second check, except it should be implemented on safekeepers, because we can't do much on pageserver - it only digests WAL that was already accepted by safekeeprers. And we need to check before approving the commit. It should be straightforward to implement because we pass feedback through safekeepers.

LizardWizzard:
> The user shouldn't be able to change it unless it has access to the VM.

Thats exactly what I'm thinking about. We cannot trust compute because it might be possible for user to escape postgres process. 

> I agree that we need a second check

Is it in the scope of this particular RFC? If so lets mention it

lubennikovaav:
> Is it in the scope of this particular RFC? If so lets mention it

Done

stepashka:
@antons-antons proposed to also implement advance warning about the limit soon to be breached

knizhnik (commented):
From point of view it should not be responsibility of safekeeper to check any limits.
There are intended to be many different limits: 
1. Logical size (total size of al relations) limit
2. Physical size (size of pageserver respository) limit
3. PiTR limit (if PiTR interval is specified in time, not LSN large)
...
Only pageserver will be able to check all this limits.
Instead of returning back to safekeeper current logical size, I think that it should report limit violation error (if any) and "severity" of this error. There are not so many things we can do in safekeeper to handle this error:
1. Propagate this error to compute node and let walsender handle it (requires extenson of replication protocol)
2. Break connection with compute node
3. Preserve connection but reply with error for each replication message (may be except requesting state)

I think that it is very important to notify compute node about detected problem and provide "soft notification": not just abort transaction but produce warning. But as far as compute node can be hacked, we at some moment should prohibit any further updates from this compute node. And IMHO this decision should be made by pageserver, not by safekeeper, which should only retranslate its decision.

lubennikovaav:
> From point of view it should not be responsibility of safekeeper to check any limits. There are intended to be many different limits: 

Thank you. I agree that the pageserver is a more appropriate place to perform this check. I've updated the RFC text.

One more question is how to config the limit on pageserver. I think that a global setting would be OK for the start, 
but to allow different limits for different clients, we need to implement per-timeline (per tenant?) config on pageserver.
AFAIK, we don't have it now.

hlinnaka:
> > The user shouldn't be able to change it unless it has access to the VM.
> 
> Thats exactly what I'm thinking about. We cannot trust compute because it might be possible for user to escape postgres process.

Correct. However, if someone bypasses the limit in the compute node, it won't cause any other harm than using more disk space. We can tolerate that to a large extent.

In addition to this limit, we should also monitor the size per tenant with Graphana, and if a free tier tenant grows wildly larger than the limit, we will notice and do something, like send an email to the user to ask what's going on.

> > I agree that we need a second check
> 
> Is it in the scope of this particular RFC? If so lets mention it

Out of scope. Would make sense to have a second check in the page server too, but it's very low priority.

hlinnaka (commented):
we've moved all the RFCs to the main repo under docs/rfcs. Can you update this to reflect the discussion here, and what was actually implemented, and open a PR to put this into docs/rfcs, please?

lubennikovaav:
Updated RFC in main repo
https://github.com/neondatabase/neon/pull/1505

PR #13, opened 2022-01-17
014-safekeeper-gossip

kelvich:
# Safekeeper gossip

## Motivation

In some situations, safekeeper (SK) needs coordination with other SK's that serve the same tenant:

1. WAL deletion. SK needs to know what WAL was already safely replicated to delete it. Now we keep WAL indefinitely.
2. Deciding on who is sending WAL to the pageserver. Now sending SK crash may lead to a livelock where nobody sends WAL to the pageserver.
3. To enable SK to SK direct recovery without involving the compute

## Summary

Compute node has connection strings to each safekeeper. During each compute->safekeeper connection establishment, the compute node should pass down all that connection strings to each safekeeper. With that info, safekeepers may establish Postgres connections to each other and periodically send ping messages with LSN payload.

[Read more](https://github.com/zenithdb/rfcs/blob/e62d021ef27e4f3cdf2757430f630ef13233ebbd/text/014-safekeeper-gossip.md)

PR #16, opened 2022-01-19
015-storage-messaging

kelvich:
## Motivation

As in [014-safekeeper-gossip ](https://github.com/zenithdb/rfcs/pull/13)we need to solve the following problems:

* Trim WAL on safekeepers
* Decide on which SK should push WAL to the S3
* Decide on which SK should forward WAL to the pageserver
* Decide on when to shut down SK<->pageserver connection

This RFC suggests a more generic and hopefully more manageable way to address those problems. However, unlike 014-safekeeper-gossip, it does not bring us any closer to safekeeper-to-safekeeper recovery but rather unties two sets of different issues we previously wanted to solve with gossip.

Also, with this approach, we would not need "call me maybe" anymore, and the pageserver will have all the data required to understand that it needs to reconnect to another safekeeper.

## Summary

Instead of p2p gossip, let's have a centralized broker where all the storage nodes report per-timeline state. Each storage node should have a `--broker-url=1.2.3.4` CLI param.

Here I propose two ways to do that. After a lot of arguing with myself, I'm leaning towards the etcd approach. My arguments for it are in the pros/cons section. Both options require adding a Grpc client in our codebase either directly or as an etcd client's dependency.

[read more](https://github.com/zenithdb/rfcs/blob/5333469856e5924acdc9e0117e61201cdbdd8e11/015-storage-messaging.md)

### DIFF
--- a/015-storage-messaging.md
+++ b/015-storage-messaging.md
@@ -34,6 +34,11 @@ adds either etcd or console as a storage dependency
 
 ## Possible implementation: custom message broker in the console
 
+We've decided to go with an etcd approach instead of the message broker.
+
+<details closed>
+<summary>Original suggestion</summary>
+<br>
 We can add a Grpc service in the console that acts as a message broker since the console knows the addresses of all the components. The broker can ignore the payload and only redirect messages. So, for example, each safekeeper may send a message to the peering safekeepers or to the pageserver responsible for a given timeline.
 
 Message format could be `{sender, destination, payload}`.
@@ -85,8 +90,10 @@ sk_1 <-> |                | <-> pserver_1
 ...      | Console broker |     ...
 sk_n <-> |________________| <-> pserver_m
 ```
+</details>
 
-## Proposed implementation: etcd state store
+
+## Implementation: etcd state store
 
 Alternatively, we can set up `etcd` and maintain the following data structure in it:
 
@@ -112,6 +119,32 @@ As etcd doesn't support field updates in the nested objects that translates to t
 ```
 
 Each storage node can subscribe to the relevant sets of keys and maintain a local view of that structure. So in terms of the data flow, everything is the same as in the previous approach. Still, we can avoid implementing the message broker and prevent runtime storage dependency on a console.
+
+### Safekeeper address discovery
+
+During the startup safekeeper should publish the address he is listening on as the part of `{"sk_#{sk_id}" => ip_address}`. Then the pageserver can resolve `sk_#{sk_id}` to the actual address. This way it would work both locally and in the cloud setup. Safekeeper should have `--advertised-address` CLI option so that we can listen on e.g. 0.0.0.0 but advertize something more useful.
+
+### Safekeeper behavior
+
+For each timeline safekeeper periodically broadcasts `compute_#{tenant}_#{timeline}/safekeepers/sk_#{sk_id}/*` fields. It subscribes to changes of `compute_#{tenant}_#{timeline}` -- that way safekeeper will have an information about peering safekeepers.
+That amount of information is enough to properly trim WAL. To decide on who is pushing the data to S3 safekeeper may use etcd leases or broadcast a timestamp and hence track who is alive.
+
+### Pageserver behavior
+
+Pageserver subscribes to `compute_#{tenant}_#{timeline}` for each tenant it owns. With that info, pageserver can maintain a view of the safekeepers state, connect to a random one, and detect the moments (e.g., one the safekeepers is not making progress or down) when it needs to reconnect to another safekeeper. Pageserver should resolve exact IP addresses through the console, e.g., exchange `#sk_#{sk_id}` to `4.5.6.7:6400`.
+
+Pageserver connection to the safekeeper can be triggered by the state change `compute_connected: false -> true`. With that, we don't need "call me maybe" anymore.
+
+As an alternative to compute_connected, we can track timestamp of the latest message arrived to safekeeper from compute. Usually compute broadcasts KeepAlive to all safekeepers every second, so it'll be updated every second when connection is ok. Then the connection can be considered down when this timestamp isn't updated for a several seconds.
+
+This will help to faster detect issues with safekeeper (and switch to another) in the following cases:
+
+      when compute failed but TCP connection stays alive until timeout (usually about a minute)
+      when safekeeper failed and didn't set compute_connected to false
+
+Another way to deal with [2] is to process (write_lsn, commit_lsn, compute_connected) as a KeepAlive on the pageserver side and detect issues when sk_id don't send anything for some time. This way is fully compliant to this RFC.
+
+Also, we don't have a "peer address amnesia" problem as in the gossip approach (with gossip, after a simultaneous reboot, safekeepers wouldn't know each other addresses until the next compute connection).
 
 ### Interactions
 
--- a/015-storage-messaging.md
+++ b/015-storage-messaging.md
@@ -1,0 +1,146 @@
+# Storage messaging
+
+Created on 19.01.22
+
+That it is an alternative to (014-safekeeper-gossip)[]
+
+## Motivation
+
+As in 014-safekeeper-gossip we need to solve the following problems:
+
+* Trim WAL on safekeepers
+* Decide on which SK should push WAL to the S3
+* Decide on which SK should forward WAL to the pageserver
+* Decide on when to shut down SK<->pageserver connection
+
+This RFC suggests a more generic and hopefully more manageable way to address those problems. However, unlike 014-safekeeper-gossip, it does not bring us any closer to safekeeper-to-safekeeper recovery but rather unties two sets of different issues we previously wanted to solve with gossip.
+
+Also, with this approach, we would not need "call me maybe" anymore, and the pageserver will have all the data required to understand that it needs to reconnect to another safekeeper.
+
+## Summary
+
+Instead of p2p gossip, let's have a centralized broker where all the storage nodes report per-timeline state. Each storage node should have a `--broker-url=1.2.3.4` CLI param.
+
+Here I propose two ways to do that. After a lot of arguing with myself, I'm leaning towards the etcd approach. My arguments for it are in the pros/cons section. Both options require adding a Grpc client in our codebase either directly or as an etcd dependency.
+
+## Non-goals
+
+That RFC does *not* suggest moving the compute to pageserver and compute to safekeeper mappings out of the console. The console is still the only place in the cluster responsible for the persistency of that info. So I'm implying that each pageserver and safekeeper exactly knows what timelines he serves, as it currently is. We need some mechanism for a new pageserver to discover mapping info, but that is out of the scope of this RFC.
+
+## Impacted components
+
+pageserver, safekeeper
+adds either etcd or console as a storage dependency
+
+## Possible implementation: custom message broker in the console
+
+We can add a Grpc service in the console that acts as a message broker since the console knows the addresses of all the components. The broker can ignore the payload and only redirect messages. So, for example, each safekeeper may send a message to the peering safekeepers or to the pageserver responsible for a given timeline.
+
+Message format could be `{sender, destination, payload}`.
+
+The destination is either:
+1. `sk_#{tenant}_#{timeline}` -- to be broadcasted on all safekeepers, responsible for that timeline, or
+2. `pserver_#{tenant}_#{timeline}` -- to be broadcasted on all pageservers, responsible for that timeline
+
+Sender is either:
+1. `sk_#{sk_id}`, or
+2. `pserver_#{pserver_id}`
+
+I can think of the following behavior to address our original problems:
+
+* WAL trimming
+  Each safekeeper periodically broadcasts `(write_lsn, commit_lsn)` to all peering (peering == responsible for that timeline) safekeepers
+
+* Decide on which SK should push WAL to the S3
+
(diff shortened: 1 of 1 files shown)


015-storage-messaging.md
```diff
@@ -0,0 +1,146 @@
...
+
+  Each safekeeper periodically sends (write_lsn, commit_lsn, compute_connected) to the relevant pageservers. With that info, pageserver can maintain a view of the safekeepers state, connect to a random one, and detect the moments (e.g., one the safekeepers is not making progress or down) when it needs to reconnect to another safekeeper. Pageserver should resolve exact IP addresses through the console, e.g., exchange `#sk_#{sk_id}` to `4.5.6.7:6400`.
+
+  Pageserver connection to the safekeeper triggered by the state change `compute_connected: false -> true`. With that, we don't need "call me maybe" anymore.
+
+  Also, we don't have a "peer address amnesia" problem as in the gossip approach (with gossip, after a simultaneous reboot, safekeepers wouldn't know each other addresses until the next compute connection).
+
+* Decide on when to shutdown sk<->pageserver connection
+
+  Again, pageserver would have all the info to understand when to shut down the safekeeper connection.
```
LizardWizzard, line 73: Am I correct that with this approach pageserver initiates connection to safekeeper all the time? Will there be some delay between compute start and this fact reaching pageserver? If safekeeper is supposed to start replication it should know whether it should retry the connection if pageserver disconnects.

kelvich: Yeah, pageserver always initiates connection in that model. If safekeeper will advertise that state change (compute_connected: false -> true) right away without waiting for ping_period to expire then delay would be minimal. It may be even a bit faster then the current approach since we are using existing connections and avoiding tcp's triple handshake.

LizardWizzard:
I agree with the approach of having separate metadata coordination service. 

I think it is better to have separate service to manage this state. There are at least two reasons. 

One that you've already mentioned is fault tolerance. When storage and console are tied this way we should make console as available as any other infrastructure services (like for example s3). So in my opinion storage should avoid having hard dependency on console.

The other reason is that users who run zenith themselves wouldn't have our console available.

So I think approach with separate metadata service is preferable. etcd is fine, but I think it is worth mentioning/discussing any alternatives, e.g consul

015-storage-messaging.md
```diff
@@ -0,0 +1,146 @@
...
+2. `pserver_#{tenant}_#{timeline}` -- to be broadcasted on all pageservers, responsible for that timeline
+
+Sender is either:
+1. `sk_#{sk_id}`, or
+2. `pserver_#{pserver_id}`
+
+I can think of the following behavior to address our original problems:
+
+* WAL trimming
+  Each safekeeper periodically broadcasts `(write_lsn, commit_lsn)` to all peering (peering == responsible for that timeline) safekeepers
```
petuhovskiy, line 57: Can we trim WAL which is not yet persisted in S3? Maybe we should also track `sk_s3_lsn` and don't trim WAL until it's uploaded to S3, or `write_lsn` includes this condition already?

015-storage-messaging.md
```diff
@@ -0,0 +1,146 @@
...
+
+* Decide on which SK should push WAL to the S3
+
+  Each safekeeper periodically broadcasts `i_am_alive_#{current_timestamp}` message to all peering safekeepers. That way, safekeepers may maintain the vector of alive peers (loose one, with false negatives). Alive safekeeper with the minimal id pushes data to S3.
```
petuhovskiy, line 61: Safekeeper with the minimal id can have bad connection to the compute (lag behind other safekeepers) or don't receive new WAL at all. Should we do anything about it?

arssher:
I agree that in cloud some way of centralized config broadcast from the console is a good thing, however I'm quite hesitant adding etcd or whatever as one more hard dependency, required in tests and local setups. What do you think about drawing the line differently or combining the approaches and using centralized service only for console-originated info and service discovery, that is, for
1. Info on which servers which timelines reside;
2. Addresses of these servers;
3. Configuration (when we start wanting changing it on the fly);

?

So, safekeepers would learn their timelines and peers for these timelines from the broker and then gossip directly. Fully centralized service allows to avoid managing these connections, but OTOH we get more decoupled system (e.g. in case of problems you have to mainly check 2 places, not 3) and possibility of running tests/locally without broker at all.

I still think that the best way of safekeeper recovery is direct WAL fetching from peers, so we need to know them anyway.

As for callmemaybe, we can rework it close to what is proposed here: safekeepers would know pageserver(s) for each of their timelines from the broker and would continiously push data for active timelines to pageservers, providing them with info about LSNs and such, allowing them to choose from whom to pull and inviting to do so. Safekeepers here are the initiators, as they learn about compute appearance first.

I don't think n-to-n connections here is a big problem as we don't normally (unless in recovery) carry WAL in such horizontal way, and gossiped info will be multiplexed into single physical connection. I mean, yes, we still have n of them instead of 1, but n won't be too big in short-mid term (even 10k async conns won't generate tremendous load), and if it becomes so, we can indeed assign safekeepers more smartly (which is actually good for load balancing anyway, I think, as the load of each tli on all safekeepers is the same. It also provides logical grouping, e.g. it would be immediately clear whether we have problems with whole safekeepers or individual timelines).

Also, though rfc explicitly says it is not about service discovery -- and by that I mean learning which timelines reside on which safekeepers/pageserver, and their addresses -- I feel it would be very weird and duplicating to have proposed broker and not use it for SD. We need to (reliably, i.e. with retries) broadcast console-originated tli assignment and config to server nodes anyway.

I'm fine with etcd. It is somewhat strange choice given that we don't need persistency (there might be related performance issues btw), but anyway. Really we need something highly available, not strongly consistent, with get/put + pub/sub by prefix interface.

015-storage-messaging.md
```diff
@@ -0,0 +1,146 @@
...
+
+* Decide on which SK should forward WAL to the pageserver
+
+  Each safekeeper periodically sends (write_lsn, commit_lsn, compute_connected) to the relevant pageservers. With that info, pageserver can maintain a view of the safekeepers state, connect to a random one, and detect the moments (e.g., one the safekeepers is not making progress or down) when it needs to reconnect to another safekeeper. Pageserver should resolve exact IP addresses through the console, e.g., exchange `#sk_#{sk_id}` to `4.5.6.7:6400`.
```
petuhovskiy, line 65: As an alternative to compute_connected, we can track timestamp of the latest message arrived to safekeeper from compute. Usually compute broadcasts KeepAlive to all safekeepers every second, so it'll be updated every second when connection is ok. Then the connection can be considered down when this timestamp isn't updated for a several seconds.

This will help to faster detect issues with safekeeper (and switch to another) in the following cases:
1. when compute failed but TCP connection stays alive until timeout (usually about a minute)
2. when safekeeper failed and didn't set compute_connected to false

Another way to deal with [2] is to process (write_lsn, commit_lsn, compute_connected) as a KeepAlive on the pageserver side and detect issues when sk_id don't send anything for some time. This way is fully compliant to this RFC.

petuhovskiy (commented):
I think both centralized and gossip approaches can be useful in different cases. I also agree that centralized approach is simpler to solve our current problems, so we can start with implementing centralized metadata broker, and after that implement gossip if we need it. Then it will be possible to choose freely between both approaches.

015-storage-messaging.md
arssher: > Maybe we should also track sk_s3_lsn and don't trim WAL until it's uploaded to S3

Right.

015-storage-messaging.md
```diff
@@ -0,0 +1,146 @@
...
+* solves "call me maybe" as well
+* avoid possible N-to-N connection issues with gossip without grouping safekeepers in pre-defined triples 
+
+## Console broker vs. etcd:
+
+Initially, I wanted to avoid etcd as a dependency mostly because I've seen how painful for Clickhouse was their ZooKeeper dependency: in each chat, at each conference, people were complaining about configuration and maintenance barriers with ZooKeeper. It was that bad that ClickHouse re-implemented ZooKeeper to embed it: https://clickhouse.com/docs/en/operations/clickhouse-keeper/. 
+
+But with an etcd we are in a bit different situation:
+
+1. We don't need persistency and strong consistency guarantees for the data we store in the etcd
```
bojanserafimov, line 176: I don't feel great about running a single-node in-memory etcd because probably we'd be the first to try that in production. If we need a third-party in-memory kv store, redis would run faster and be easier to manage.

kelvich: Yeah, I don't suggest running it in a single-node setup in production. My point was that if we later want to drop etcd as a dependency for self-hosted installations, it would be easier to do so if we do not store any persistent data there -- we can change it to Redis or swap it with custom grpc service. But once we start keeping reliable info there, it would be way harder to switch to something in-house.

015-storage-messaging.md
kelvich: Not sure. We can add some reconnection heuristic if we need it.

015-storage-messaging.md
kelvich: +1

015-storage-messaging.md
```diff
@@ -0,0 +1,179 @@
...
+
+Each storage node can subscribe to the relevant sets of keys and maintain a local view of that structure. So in terms of the data flow, everything is the same as in the previous approach. Still, we can avoid implementing the message broker and prevent runtime storage dependency on a console.
+
+### Safekeeper address discovery
+
+During the startup safekeeper should publish the address he is listening on as the part of `{"sk_#{sk_id}" => ip_address}`. Then the pageserver can resolve `sk_#{sk_id}` to the actual address. This way it would work both locally and in the cloud setup. Safekeeper should have `--advertised-address` CLI option so that we can listen on e.g. 0.0.0.0 but advertize something more useful.
+
+### Safekeeper behavior
+
+For each timeline safekeeper periodically broadcasts `compute_#{tenant}_#{timeline}/safekeepers/sk_#{sk_id}/*` fields. It subscribes to changes of `compute_#{tenant}_#{timeline}` -- that way safekeeper will have an information about peering safekeepers.
```
LizardWizzard, line 129: This is not specific to safekeeper, but what is the period? And how the reader can identify the data as stale or its advertiser is down? There could be some threshold value, e.g no updates in 2 * update_period. Do we need to guard against possible clock skew?

kelvich: > what is the period?

Period would delay WAL trim / S3 pushdown switchover / pageserver reconnection time (only for stale connections, though; usually pageserver will receive a tcp error). I think 1 second is fine.

> And how the reader can identify the data as stale or its advertiser is down? 

Safekeeprs only care about median(lsn) among the group, so they can avoid that.

Pageserver needs to check that the safekeeper it is connected to is not extremly behind the peering safekeepers. We can try to do that based on the amount of LSN lag between our safekeeper and rest of them, but that approach can lead to a livelock. Alternatively, we can advertise the time of the last heartbeat from the compute in each safekeeper (right now heartbeats are sent each second) and have quite a big timeout to force a re-connection in the pageserver (let's say 30 sec) to defend against the time skew. Which should be fine since that is a last resort measure -- with most of usual the problems on the safekeeper (crash, restart, oom, etc) we should be able to detect disconnect on the pageserver and that 30 sec timeout would happen only in rare connectivity / freeze scenarios. IIRC protocol level heartbeats between safekeeper and pageserver are not enabled right now -- we can enable them also and have a smaller disconnect timeouts there (since we may just measure delay without looking at the exact value of the timestamp).

LizardWizzard: > median(lsn)

MIN for commit and MAX to calculate the lag, right?

> to defend against the time skew

I think there is one more option, we can defend even from a connection thread on the safekeeper being stuck, not only the clock skew. We can compare last heartbeat time not with the pageserver current system time, but with values advertised by other safekeepers. In that case safekeeper should forward system time from compute, avoiding generation of it's own timestamp. Then all the compared values are from the same time source. Exception to that is only when there is a different compute, which shouldn't be a problem and even if it is we can add some compute id, to avoid comparison of timestamps from different computes. What do you think?

I've tried to make some sketches, this is the version with livelock I used to imagine different scenarios: (click to open in full screen)

mermaid-diagram-20220126193556

I can fix it to use timestamp comparison and if you have some failure scenarios in mind I can add them too. I think it is a good idea to add strict sequence of actions for different kinds of failures.

The other possible diagram might describe concurrent checkpoint handling by pageservers. I'm still thinking what type of coordination we need here. Can we even avoid it somehow?

For that case pageserver which is currently streaming to s3 can advertise the flag that it is streaming, latest checkpoint lsn and a timestamp, this timestamp may be the max value of compute heartbeat timestamp it has seen. This way the other pageserver can detect the difference between the current max timestamp and the advertised one, and take over the streaming role. Thoughts?

arssher: I doubt we need to fight with clock skews here. How about the following straw man for choosing the safekeeper to fetch WAL from:
1. If there is no live connection to safekeeper, choose the one with max commit_lsn.
2. Otherwise, switch to safekeeper with higher commit_lsn iff either 1) timeout (tracked locally on pageserver) passed with no data coming through current connection (prevents livelock) 2) diff between commit_lsn of our safekeeper and max commit_lsn exceeded some threshold -- we are talking with deeply lagging safekeeper.

This procedure can be run regularly + on interesting events, like lost connection to safekeeper or got new data from the broker.

Generally tracking of live servers without bothering about time skew can be done by pushing keys to etcd with expiration period, but not sure where we need that currently.

LizardWizzard: I think it is worthwhile to avoid clock skew early on during initial system design. 

I've modified the diagram to match the picture we've agreed upon: (click to open full screen)

image

Let me know if I missed something or you spot any errors.

LizardWizzard: This is my understanding of tenant migration process with metadata service: (click to open full screen)

image

PR #17, opened 2022-02-04
LSM tree approach to storage.

hlinnaka:
Per discussion in offsite meeting.

### DIFF
--- a/storage-lsm.md
+++ b/storage-lsm.md
@@ -1,0 +1,146 @@
+# Why LSM trees?
+
+In general, an LSM tree has the nice property that random updates are
+fast, but the disk writes are sequential. When a new file is created,
+it is immutable. New files are created and old ones are deleted, but
+existing files are never modified. That fits well with storing the
+files on S3.
+
+Currently, we create a lot of small files. That is mostly a problem
+with S3, because each GET/PUT operation is expensive. Currently, the
+files "archived" together into larger checkpoint files before they're
+uploaded to S3, but garbage collecting data from the archive files
+would be difficult and we have not implemented it. This proposal
+addresses that problem.
+
+
+# Overview
+
+
+```
+^ LSN
+|
+|      Memtable:     +-----------------------------+
+|                    |                             |
+|                    +-----------------------------+
+|
+|
+|            L0:     +-----------------------------+
+|                    |                             |
+|                    +-----------------------------+
+|
+|                    +-----------------------------+
+|                    |                             |
+|                    +-----------------------------+
+|
+|                    +-----------------------------+
+|                    |                             |
+|                    +-----------------------------+
+|
+|                    +-----------------------------+
+|                    |                             |
+|                    +-----------------------------+
+|
+|
+|           L1:      +-------+ +-----+ +--+  +-+
+|                    |       | |     | |  |  | |
+|                    |       | |     | |  |  | |
+|                    +-------+ +-----+ +--+  +-+
+|
+|                       +----+ +-----+ +--+  +----+
+|                       |    | |     | |  |  |    |
+|                       |    | |     | |  |  |    |
+|                       +----+ +-----+ +--+  +----+
+|
++--------------------------------------------------------------> Page ID
+
+
++---+
+|   |   Layer file
++---+
+```
+
+
+# Memtable
+
+When new WAL arrives, it is first put into the Memtable. Despite the
+name, the Memtable is not a purely in-memory data structure. It can
+spill to a temporary file on disk if the system is low on memory, and
+is accessed through a buffer cache.
+
+If the page server crashes, the Memtable is lost. It is rebuilt by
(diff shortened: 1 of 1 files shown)


storage-lsm.md
```diff
@@ -0,0 +1,146 @@
...
+
+# Next steps
+
+- Allow delta layers to cover a range keys instead of a single segment.
+
+- Implement a two-level LSM tree (or three-leveled, if you count the
+"memtable"), by adding L0.
+
+# Additional Reading
+
```
knizhnik, line 144: Although I agree that segments with varying boundaries (containing multiple relations) seems to be the only possible choice if we want to avoidlarger number of small files, it may significantly complicate GC and checkpointing.
Right now checkpointer produces both image and delta layers. It is not good and cause huge write amplification.

In my PR I reduced number of image layers based on number of intermediate delta layers and their total size.
The idea is that  we should not produce image layer if there were not so much changes, but we should not create too creating of too larger number of delta layers is also not good because cause read amplification: we may need to read multiple delta layers to reconstruct page.

It seems to be non trivial to find right criteria for producing image layers.
With new approach this task is even more complicated because layers may overlap.
Assume that we have to table T1 and T2, where T1is actively updated an T2 - not.
We create delta layer L1 containing both T1 and T2 changes. Then we change mostly T1 and produce a lot of delta layers for T1 . GC can not remove L1 until image layers both for T1 and T2 were produced.

storage-lsm.md
```diff
@@ -0,0 +1,146 @@
...
+When new WAL arrives, it is first put into the Memtable. Despite the
+name, the Memtable is not a purely in-memory data structure. It can
+spill to a temporary file on disk if the system is low on memory, and
+is accessed through a buffer cache.
+
+If the page server crashes, the Memtable is lost. It is rebuilt by
+processing again the WAL that's newer than the latest layer in L0.
+
+The size of the Memtable is equal to the "checkpoint distance", or the
+amount of WAL that we need to keep in the safekeeper.
```
kelvich, line 75: Would be good to offload remaining WAL for idle clusters to S3 as well. Which probably means that we need to offload the memtable content after some timeout and bump s3_consistent_lsn to allow safekeeper to delete that WAL.

storage-lsm.md
```diff
@@ -0,0 +1,146 @@
...
+# Partitioning scheme
+
+When compaction happens and creates a new set of files in L1, how do
+we partition the data into the files? 
+
+- Goal is that each file is ~ 1 GB in size
+- Try to match partition boundaries at relation boundaries. (See [1]
+  for how PebblesDB does this, and for why that's important)
+- Greedy algorithm
+
```
MMeent, line 135: Are the list items above seperate answers to the question of partitioning, or requirements for the partitioning algorithm? 

More specifically; putting a goal of 1GB per file is possible, but (as Konstantin mentions as well) significantly increases the complexity of GC due to not easily being able to find the files that apply to a certain page.

Splitting the layer compaction on key range (e.g. `hash(PageId) % n_buckets`) works well in reducing update overhead and spreading update skew over the key range buckets.
This seems especially important when we want to implement sharding; as the shard boundaries can now be trivially implemented on these LSM bucket boundaries; helping tremendously in reducing any download overhead for newly created shards that need to download the data they are responsible for from the original or S3. 

Lastly, splitting at 1GB boundaries of every X GB is not significantly different from just merging those X GB in a single file, other than that it now consumes X times as many file descriptors. I don't think that's a great idea overall.

storage-lsm.md
bojanserafimov: I think [this](https://github.com/zenithdb/zenith/issues/1142) addresses the problem. Here's what would happen:
1. We write delta layer L1 with changes to T1 and T2
2. We write lots of deltas on T1, and we make sure these deltas also contain a page image for each page modified in L1
3. We delete L1

There could be some simpler short-term solution, but eventually we'll need this for read performance anyway.

storage-lsm.md
hlinnaka: They are goals, or "squishy requirements" :-). It's true that we cannot both respect relation boundaries strictly, and have 1 GB physical files. We should have some kind of a heuristic that compromises between the two goals. For example, look at the layer below, and try to use the same partitioning. If it would lead to some partitions that are < 0.5 GB, merge them together, or if one partition would be larger than 2 GB, split it.

"Greedy algorithm" is a suggestion on how the implementation might work, not a goal or requirement as such.

hlinnaka:
I created another presentation to accompany this RFC:

https://docs.google.com/presentation/d/1FugYvzAJglS11YtsrcO4mbPPFSBEhsTwg931-uQlxRk/edit?usp=sharing

hlinnaka:
This has been moved to be part of PR [zenith/zenithdb#](https://github.com/zenithdb/zenith/pull/1350).

PR #18, opened 2022-04-01
Reduction of the GetPage@LSN problem to GetPage@LatestLSN

bojanserafimov:


text/014-get-page-recuction.md
```diff
@@ -0,0 +1,52 @@
...
+Statement: The pageserver will only get two kinds of get_page queries:
+a) get a hot page at some LSN within `hot_wal_size`
+b) get the latest cold version of a page
+
+Proof: It follows from the definition of `hot_wal_size` as an upper bound on replica lag. 
+
+### 4.2 Lemma 2
+Statement: Answering queries of type (a) is easy.
+
+Proof: The `hot_wal_size` fits in RAM. One implementation of (a) is to just read the `PageReconstructData` from ram, and optimize our wal redo enough so that get_page is pretty fast. Some people believe wal redo can be as fast as memcopy. Others believe that there alternative in-memory data structures that obviate the need for redo. I'm sure there are other approaches, but in the end this is easier than finding `PageReconstructData` from disk. I refrain from picking a solution in order to avoid strawman arguments.
```
knizhnik, line 36: I do no agree with this lemme. Let wal records which has to be applied fit in RAM. But them has to be applied to some page image. And this page may not be present in RAM. So we have to read if from disk. And it can be quite expensive.

bojanserafimov: Read the page image from disk, read the wal from ram. Reading 8kb from disk is way faster than reading megabytes of scattered PageReconstructData from disk.

MMeent: I also don't agree with this lemma. The definitions used in (3) apply to safekeepers; not necessarily PageServers. PageServers need to index the received WAL and thus might need potentially significant overhead for this index. Therefore, the size of the WAL at the safekeeper is not equivalent to the size of the WAL at the PageServer.

Furthermore; that something fits in RAM doesn't mean that it is easily accessed. Finding an arbitrary value in the digits of Pi is not somehow trivializable, even if all digits of Pi were somehow pre-loaded into RAM: the operation still is O(n); and O(n) on the example's 80GB of data is not something to trivialize with "it is easy".

Sure, linear search might be easy to implement, but it is not performant, and we care a lot about performance.

